PR #3, opened 2021-01-19
Tighten type definitions for useLocalStorage

LoneRifle:
- Specify the return type for the hook
- Pass type argument T to `useState()` so that we have typed values


Junscuzzy (approved):
Thanks a lot for your contribution! :)

PR #6, opened 2021-03-25
Various improvements to useInterval hooks

jonkoops:
Includes the following changes
- `savedCallback` uses the callback as it's default value, removing the need to check for undefined values and allowing it's type to be inferred
- `savedCallback` will only have it's value updated if a new callback was specified (e.g. upstream memoized values will not be re-assigned)
- A delay with a null is handled in an early return, reducing nesting of code.

### DIFF
--- a/src/hooks/useInterval/useInterval.ts
+++ b/src/hooks/useInterval/useInterval.ts
@@ -1,26 +1,25 @@
 import { useRef, useEffect } from 'react'
 
-function useInterval(callback: () => void, delay: number | null) {
-  const savedCallback = useRef<() => void | null>()
+export default function useInterval(
+  callback: () => void,
+  delay: number | null,
+) {
+  const savedCallback = useRef(callback)
 
-  // Remember the latest callback.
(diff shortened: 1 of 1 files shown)


src/hooks/useInterval/useInterval.ts
```diff
@@ -1,26 +1,25 @@
...
+  delay: number | null,
+) {
+  const savedCallback = useRef(callback)
 
-  // Remember the latest callback.
+  // Remember the latest callback if it changes.
   useEffect(() => {
     savedCallback.current = callback
-  })
+  }, [callback])
```
jonkoops, line 12: I am wondering if it's the right approach to wrap this in a `useEffect` at all, it might be more performant to simply re-assign this value on every render.

Junscuzzy (approved):
Hi @jonkoops, thanks you for you contribution, your suggest is great, thanks a lot! :100:

Junscuzzy:
It's in production: https://usehooks-typescript.com/react-hook/use-interval 
Thanks :)

PR #9, opened 2021-05-12
Even stronger type for eventName in useEventListener.ts

Nishit-Dua:
Hi, so string might be too lose of a type for this, since we could put some random string as an event.

`useEventListener("random", ()=>{})` didn't catch that "random" isn't an event listener type, so i updated it.

### DIFF
--- a/generated/hooks/useEventListener.hook.md
+++ b/generated/hooks/useEventListener.hook.md
@@ -1,7 +1,8 @@
-```tsimport { useRef, useEffect, RefObject } from 'react'
+```ts
+import { useRef, useEffect, RefObject } from 'react'
 
 function useEventListener<T extends HTMLElement = HTMLDivElement>(
-  eventName: string,
+  eventName: keyof WindowEventMap,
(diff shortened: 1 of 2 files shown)


generated/hooks/useEventListener.hook.md
```diff
@@ -1,7 +1,8 @@
-```tsimport { useRef, useEffect, RefObject } from 'react'
+```ts
+import { useRef, useEffect, RefObject } from 'react'
 
 function useEventListener<T extends HTMLElement = HTMLDivElement>(
-  eventName: string,
+  eventName: keyof WindowEventMap,
```
Junscuzzy, line 5: Thank you to think update this here, but don't worries, the `generated/**/*.md` is generated by script. 

You can see this in the package.json, it's named `"copy:hooks"`

Junscuzzy:
Hi @Nishit-Dua, thank you for the contribution!

`WindowEventMap` is very better than `string` :100:

PR #12, opened 2021-05-26
Update on setState to make it safe on change.

joaoskr:
### As setState is async, we may have sideeffect problems as we can see in the example below.

https://codesandbox.io/s/setstatesafe-62pmg


Junscuzzy:
Hi @joaoskr, oh yeah good catch! Thank you to your contribution, I will include that change very sooner, thank dude!

PR #25, opened 2021-08-23
Add return type

datkira:
I think we should add a return type to this function so that the warning in typescript-eslint does not appear.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md
Warning message: "ESLint: Missing return type on function.(@typescript-eslint/explicit-module-boundary-types)"


juliencrn:
Good catch! Thank you @datkira 👌

PR #38, opened 2021-10-05
Slightly better wording for `useCopyToClipboard`

n1c:
This could also have just been changed to "gives you" but "provides" feels nicer?


juliencrn (approved):
Yeah! You're true, thanks @n1c! 👍

PR #45, opened 2021-10-28
useElementSize should use useLayoutEffect

jaslong:
Many apps need to synchronously grab the element size to avoid a flicker where the element size is still 0. This changes solves the flicker.


juliencrn:
Thanks @jaslong 👍

PR #48, opened 2021-11-05
feat(hooks): add useStep

qlaffont:
Able to have a simple stepper function.


lib/src/useStep/useStep.ts
```diff
@@ -0,0 +1,62 @@
...
+
+interface ReturnType {
+  actualStep: number
+  goToNextStep: () => void
+  goToPrevStep: () => void
+  reset: () => void
+  canGoToNextStep: boolean
+  canGoToPrevStep: boolean
+  setStep: (step: number) => void
+}
```
juliencrn, line 11: I think returning a tuple is more conventional for a hook, like this: 

```suggestion
interface Methods {
  goToNextStep: () => void
  goToPrevStep: () => void
  reset: () => void
  canGoToNextStep: boolean
  canGoToPrevStep: boolean
  setStep: (step: number) => void
}

type ReturnType = [number, Methods]
```

lib/src/useStep/useStep.ts
```diff
@@ -0,0 +1,62 @@
...
+
+function useStep(maxStep: number): ReturnType {
+  const [actualStep, setActualStep] = useState(1)
```
juliencrn, line 14: Not very important, but I think that "current" is better than "actual".

```suggestion
const [currentStep, setCurrentStep] = useState(1)
```

lib/src/useStep/useStep.ts
```diff
@@ -0,0 +1,62 @@
...
+  const canGoToPrevStep = useMemo(() => actualStep - 1 >= 1, [actualStep])
+
+  const setStep = useCallback(
+    (step: number) => {
+      if (step >= 1 && step <= maxStep) {
+        setActualStep(step)
+        return
+      }
+
+      throw 'Step not valid'
```
juliencrn, line 30:
```suggestion
      throw new Error('Step not valid')
```

lib/src/useStep/useStep.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  return {
+    actualStep,
+    goToNextStep,
+    goToPrevStep,
+    canGoToNextStep,
+    canGoToPrevStep,
+    setStep,
+    reset,
+  }
```
juliencrn, line 59: Related to previous comments, if we change the return type and the "actual" word, have to update this too.

site/src/hooks-doc/useStep/useStep.demo.tsx
```diff
@@ -0,0 +1,27 @@
...
+export default function Component() {
+  const {
+    actualStep,
+    canGoToPrevStep,
+    canGoToNextStep,
+    goToNextStep,
+    goToPrevStep,
+    reset,
+    setStep,
+  } = useStep(5)
```
juliencrn, line 14: Related to previous comments, if we change the return type and the "actual" word, have to update this too.

juliencrn (changes requested):
Hi @qlaffont,

Thanks a lot for this amazing contribution, well written!
This is the first hook completely created (hook, demo, test, markdown) by the community! 🎉 

I've added some little review comments, feel free to fix it by yourself or if you prefer, can do it too.

Last thing, have you encountered some frustration in the hook creation process?

Thanks again, Quentin 👍

PR #59, opened 2021-11-09
Fixed edit link

ducktordanny:
# Fixed edit link on site

## Bug

I just found this repository through an article what I found on [daily.dev](https://daily.dev) and when I wanted to open a hook's github page from the site (at: `See a way to make this page better? Edit there »`) I got a 404.

## Fix

So after that I took a look at the repository and I found out that the link was pointing at a branch where the folders wasn't available already I looked for the hooks in the master branch. I replaced the editLink's path in the `site/src/templates/post.tsx` and it worked correctly again.

Great library! :)


juliencrn:
Hey @ducktordanny, good catch! Thanks 👏

PR #61, opened 2021-11-10
useElementSize doesn't update size when elementRef.current changes

### LINKED ISSUE #57: useElementSize doesn't update size when elementRef.current changes
# Summary

If `elementRef.current` within `useElementSize` starts out as `null` on initial render (typically because the target element does not render), `useElementSize` will not update and return the correct element size once `elementRef.current` changes to a non-null value.

# Simplified code example

```ts
const FullScreenDialog = ({ open, title, msg }: { open: boolean; title: string; msg: string }) => {
    const hdrRef = useRef(null);
    const { height: winHeight } = useWindowSize();
    const { height: hdrHeight } = useElementSize(hdrRef);

    if (!open) return null;
    return (
        <div>
            <div ref={hdrRef}>{title}</div>
            <div style={{ height: winHeight - hdrHeight, marginTop: hdrHeight }}>{msg}</div>
        </div>
    );
};
```

# Details

I'm using `useElementSize` in a full screen dialog component similar to the simplified one above. The dialog component starts out as not open (`open=false`), so on first render, it returns `null`. This means that the dialog header element I want the size of doesn't render at all on initial render and therefore, `elementRef.current` is `null`. The `useLayoutEffect` within `useElementSize` runs and calls `handleSize`, but because `elementRef.current` is `null`, `setSize` does not get called and so `useElementSize` returns `{ width: 0, height: 0 }`. This is all working as expected so far.

Then, the user clicks a button to open the dialog. So my dialog component re-renders, but now with `open=true` so it renders to the actual elements of the dialog component. Now `elementRef.current` becomes an actual node. However, there is nothing that causes `handleSize` to get called again, so `useElementSize` will always return `{ width: 0, height: 0 }`. (Note: resizing the window does cause `handleSize` to get called and then `useElementSize` returns the correct element size.)

# Suggested fix

I believe the fix is simple: add `elementRef?.current` to the item list in the call to `useLayoutEffect`. Then, on a subsequent re-render of `useElementSize`, when `elementRef?.current` changes from `null` to non-null, `handleSize` will get called and the size will be updated.

# Other info

This could possibly be related to #32.

mterrel:
Fixes #57

### DIFF
--- a/lib/src/useElementSize/useElementSize.ts
+++ b/lib/src/useElementSize/useElementSize.ts
@@ -15,25 +15,25 @@ function useElementSize<T extends HTMLElement = HTMLDivElement>(
     width: 0,
     height: 0,
   })
+  const node = elementRef?.current
 
   // Prevent too many rendering using useCallback
   const handleSize = useCallback(() => {
-    const node = elementRef?.current
     if (node) {
       setSize({
         width: node.offsetWidth || 0,
         height: node.offsetHeight || 0,
       })
     }
-  }, [elementRef])
+  }, [node])
 
   useEventListener('resize', handleSize)
 
-  // Initial size on mount
+  // Initial size on mount and when current element changes
   useLayoutEffect(() => {
     handleSize()
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
+  }, [node])
 
   return size
 }


juliencrn:
Thanks for your contribution!

After some tests with the [demo component](https://usehooks-ts.com/react-hook/use-element-size), it seems not work as expected. 
Indeed, on the demo example, I have `{ width: 0, height: 0 }` in the first render, have to update it to get the size.

I think is because we can't use the reference in the dependency list, it's a mutable value and don't update the component. It was the case, and it's why it doesn't work well.

> React Hook useEffect has an unnecessary dependency: 'ref.current'. Either exclude it or remove the dependency array. Mutable values like 'ref.current' aren't valid dependencies because mutating them doesn't re-render the component. (react-hooks/exhaustive-deps)

https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom

The [React documentation](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) proposes another solution, but we can't listen the resize event with this method.

I tried to store the ref as a state instead, to be reactive.
But it introduces a BREAKING CHANGE with the new hook API.

I created a [CodeSandox](https://codesandbox.io/s/usehooks-ts-react-component-template-forked-3o31s?file=/src/useElementSize.ts) with a modified hook and your example. 

What do you think about?

mterrel:
@juliencrn Yes, this looks like a more complete solution. Thanks for looking more closely at it!

PR #62, opened 2021-11-16
Added useClickAnyWhere

sonjoydatta:
Hi @juliencrn, 
I just added a simple hook called "useClickAnyWhere". 
Can you review it?

Thanks.

### DIFF
--- a/packages/usehooks-ts/src/useClickAnyWhere/useClickAnyWhere.ts
+++ b/packages/usehooks-ts/src/useClickAnyWhere/useClickAnyWhere.ts
@@ -1,0 +1,15 @@
+import { useEffect } from 'react'
+
+type AnyEvent = MouseEvent | TouchEvent
+type Handler = (event: AnyEvent) => void
+
+function useClickAnyWhere(handler: Handler) {
+  useEffect(() => {
+    const callback = (event: AnyEvent) => handler(event)
+
+    document.addEventListener('click', callback)
+    return () => document.removeEventListener('click', callback)
+  }, [handler])
+}
+
+export default useClickAnyWhere
--- a/packages/usehooks-ts/src/index.ts
+++ b/packages/usehooks-ts/src/index.ts
@@ -1,4 +1,5 @@
 export * from './useBoolean'
+export * from './useClickAnyWhere'
 export * from './useCopyToClipboard'
 export * from './useCounter'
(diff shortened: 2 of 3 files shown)


packages/usehooks-ts/src/useClickAnyWhere/useClickAnyWhere.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect } from 'react'
+
+type AnyEvent = MouseEvent | TouchEvent
+type Handler = (event: AnyEvent) => void
```
juliencrn, line 4: Can be simplified to 

```suggestion
type Handler = (event: MouseEvent) => void
```

Indeed, the `TouchEvent` is keyboard related and not called in the hook.

packages/usehooks-ts/src/useClickAnyWhere/useClickAnyWhere.ts
```diff
@@ -0,0 +1,15 @@
...
+
+function useClickAnyWhere(handler: Handler) {
+  useEffect(() => {
+    const callback = (event: AnyEvent) => handler(event)
+
+    document.addEventListener('click', callback)
+    return () => document.removeEventListener('click', callback)
```
juliencrn, line 11: Here, two things:
- We can simplify the hook using the `useEventListener`.
- The Typescript types check doesn't pass, we have to type it manually. 

```suggestion
    useEventListener('click', event => {
      handler(event as unknown as MouseEvent)
    })
```

juliencrn:
Hello @sonjoydatta!

Thank you for your nice work, and good idea, we will add it 👍

There are some points to see:
- The `develop` branch is a little outdated, my bad, could be better to rebase on `master`.
- Do you also want to add the description to display it on the site? See the `npm run plop` command ;)

PS: I discover an issue in another event based hook while doing this code review, cool!

juliencrn (changes requested):

packages/usehooks-ts/src/useClickAnyWhere/useClickAnyWhere.ts
sonjoydatta: Good idea!

juliencrn:
Branch moved to #64

PR #69, opened 2021-11-30
add param to allow 'mouseup' event

JamesBarrettDev:
In some cases mousedown fires too early. For example, when clicking on a button below a collapsible section using useOnClickOutside the layout shifts causing the buttons onClick event to be missed. Having an option to change the mouse event to mouseup would help ensure both onClick and useOnClickOutside events work.


juliencrn:
Yes 👍. Thanks, @JamesBarrettDev.

thomask:
@juliencrn Is there anything against also adding "click"? Because this would allow to `stopPropagation` of the event from an onClick handler in a nested component. If not, I will create a PR.

JamesBarrettDev:
> @juliencrn Is there anything against also adding "click"? Because this would allow to `stopPropagation` of the event from an onClick handler in a nested component. If not, I will create a PR.

Probably worth adding `onDoubleClick` too so all click events are covered?

PR #77, opened 2021-12-31
More strict typing

alvaroserrrano:


juliencrn:
Hi @alvaroserrrano, thank you for your contribution! 
Good catch, the typo 👍  I've included it by cherry-picking the commit.

But I'll not merge the branch because of the second commit. 
I think you don't use the project's config (Typescript and ESLint config at least).

Anyway, thanks for your time.

PR #86, opened 2022-01-18
Feature: ternary dark mode

PabloLION:
- Add hook `useTernaryDarkMode`
- With [Demo on CodeSandbox](https://codesandbox.io/s/usehooks-ts-react-component-template-forked-lprwq?file=/src/App.tsx)
- Close #81 (for the "How to Contribute" part maybe we move to a new issue)
- Passed `npm run test`

PR #89, opened 2022-01-20
Feature/use ternary dark mode

### LINKED ISSUE #81: Can I PR new functions?
Hi team. 
I made a ternary Darkmode (based on this repo) like `"dark" | "system" | "light"` in a project  and I want to use it in another project.  
The best way to share it is to upload it on NPM and GitHub. And I think putting it here might be helpful, as I see you have " Add new hooks" in your roadmap. Questions are:

1. Is this too trivial to add?
2. Can I make a PR on this repo? 
3. Do you have a Code of Conduct? 

Thanks.

PabloLION:
I don't know how to fix my rebase on PR #86 .. So I created a new branch.
This is the latest update (repeating PR #86)

About the SSR problem, I'm going to just apply `useMediaQuery`, since the `useDarkMode` is using it as well. We will only need to change one place to fix all.

---

The design was for using one button for the minimalist style. You can check it on the top bar of https://www.snippethub.dev/ to see how it works. For me this is better than using an extra button. It keeps the dark/light toggle language and adds a new function. This is how I thought when I wrote the function.

If we need another toggle, it's better to use the "use dark mode" and a "use system color override".

---

The other problem @juliencrn  shows in the video is fixed.

Git workflow: this close #81


lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -0,0 +1,65 @@
...
+    toggleTernaryDarkMode: () => {
+      setTernaryDarkMode((currMode: TernaryDarkMode) =>
+        currMode === TernaryDarkMode.Dark
+          ? TernaryDarkMode.Light
+          : currMode === TernaryDarkMode.Light
+          ? TernaryDarkMode.System
+          : TernaryDarkMode.Dark,
+      )
+    },
+  }
```
juliencrn, line 62: - Please add a method like `setTernaryDarkMode` to allow more hook implementations.
- For `toggleTernaryDarkMode`, it's quite better to create a function in the hook body and pass it to the return statement. Why? For clarity and because this pattern could create memory leak because of the anonymous function that cannot be collected by the garbage collector.

```ts
const toggleTernaryDarkMode = () => {
  // -- snip --
}

const setTernaryDarkMode = () => {
  // -- snip --
}

return {
  // -- snip -- 
  toggleTernaryDarkMode,
  setTernaryDarkMode
}

File: site/src/hooks-doc/useTernaryDarkMode/useTernaryDarkMode.demo.tsx [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+    <div>
+      <p>Current theme: {isDarkMode ? 'dark' : 'light'}</p>
+      <p>ternaryMode: {ternaryDarkMode}</p>
+      <button onClick={toggleTernaryDarkMode}>
+        Toggle from {ternaryDarkMode}
+      </button>
+      <p>
+        enum strict equal test:{' '}
+        {ternaryDarkMode === 'system' ? 'using System' : 'not using System'}
+      </p>
```
juliencrn, line 19: Add an example with the new method `setTernaryDarkMode` like a simple `<button>`

site/src/hooks-doc/useTernaryDarkMode/useTernaryDarkMode.mdx
```diff
@@ -0,0 +1,12 @@
...
+title: useTernaryDarkMode
+date: '2022-01-11'
+---
+
+This React Hook offers you an interface to toggle and read the dark theme mode between three values.
+
+The returned value `isDarkMode` is a boolean to let you be able to use with your logic.
+The returned value `ternaryModeCode` is a enum of string of type `"dark" | "system" | "light"`.
+When `ternaryModeCode` is set to `system`, the `isDarkMode` will use system theme, like in iOS and MacOS.
+
```
juliencrn, line 11: Add a sentence to talk about the new `setTernaryDarkMode` method.

juliencrn (changes requested):
Hi @PabloLION,

Thanks for your new PR!

Don't worry about the rebase, I've tested locally, there isn't any conflict, we'll be able to do it with the GitHub UI (rebase and merge).

About the design, I still think that the toggle method is a bit too specific for a library like this, but maybe I'm wrong! We'll keep it.
But, imagine an HTML `<select>` to choose the ternary theme, we can't handle it with our current hook. 
To solve it, we could add a more low-level method (like just a wrapped `setTernaryDarkMode`) to allow other implementations.

I was happy to see an `enum`, but for the new functionality, it could be tricky or redundant to implement. Feel free to come back to a simpler type like `type TernaryDarkMode = 'system' | 'dark' | 'light'`, more convenient to be exposed as the public API.

Some stuff could help:
- `setState<T>: React.Dispatch<React.SetStateAction<T>>`
- `type TernaryDarkModeKeys = keyof typeof TernaryDarkMode`

lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -0,0 +1,65 @@
+import { useEffect, useState } from 'react'
+
+// See: https://usehooks-ts.com/react-hook/use-local-storage
+import { useLocalStorage } from '../useLocalStorage'
+// See: https://usehooks-ts.com/react-hook/use-media-query
+import { useMediaQuery } from '../useMediaQuery'
+
+import { useUpdateEffect } from '..'
```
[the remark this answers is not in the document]
PabloLION: Yes, It's in the `useDarkMode`.

lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -0,0 +1,65 @@
...
+
+enum TernaryDarkMode {
+  System = 'system',
+  Dark = 'dark',
+  Light = 'light',
+}
+interface UseTernaryDarkModeOutput {
+  isDarkMode: boolean
+  ternaryDarkMode: TernaryDarkMode
+  toggleTernaryDarkMode: () => void
```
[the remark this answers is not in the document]
PabloLION: @juliencrn Good idea!

lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function useTernaryDarkMode(): UseTernaryDarkModeOutput {
+  const isDarkOS = useMediaQuery(COLOR_SCHEME_QUERY)
+  const [ternaryDarkMode, setTernaryDarkMode] =
+    useLocalStorage<TernaryDarkMode>('ternaryDarkMode', TernaryDarkMode.System)
+  const [isDarkMode, setDarkMode] = useState<boolean>(isDarkOS)
```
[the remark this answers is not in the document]
PabloLION: Yes. This part was written when I used `type TernaryDarkMode = 'system' | 'dark' | 'light'`. I admit that this enum is unnecessary, I'm switching back.

lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
PabloLION: I didn't know about the memory leak. Maybe it would be a problem. Using a `map` to do it now.

PabloLION:
```TS
      const toggleMap = new Map<TernaryDarkMode, TernaryDarkMode>([
        ['light', 'system'],
        ['system', 'dark'],
        ['dark', 'light'],
      ])
      setTernaryDarkMode((tdm: TernaryDarkMode) => toggleMap.get(tdm))
```

I can't fix the problem with map. TS deduced it's a `TernaryDarkMode | undefined`. I'm going back with `object` and `Record<>`

site/src/hooks-doc/useTernaryDarkMode/useTernaryDarkMode.demo.tsx
PabloLION: added on `da06b5c`

site/src/hooks-doc/useTernaryDarkMode/useTernaryDarkMode.mdx
PabloLION: added on `b553466`

PabloLION:
Hi @juliencrn. I agree on all points you mentioned and make changes on them. 
Please the latest commits. 
Also, thanks for teaching me how to design a robust hook.

lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -9,56 +9,63 @@ import { useUpdateEffect } from '../useUpdateEffect'
...
-enum TernaryDarkMode {
-  System = 'system',
-  Dark = 'dark',
-  Light = 'light',
-}
+type TernaryDarkMode = 'system' | 'dark' | 'light'
 interface UseTernaryDarkModeOutput {
   isDarkMode: boolean
   ternaryDarkMode: TernaryDarkMode
+  dispatchTernaryDarkMode: (ternaryDarkMode?: TernaryDarkMode) => void
```
juliencrn, line 16: Here, there is a good practice, we can return the `dispatchTernaryDarkMode` with the same API of `setState` from `useState`.

```js
import { Dispatch, SetStateAction } from 'react'
```

```suggestion
dispatchTernaryDarkMode: Dispatch<SetStateAction<TernaryDarkMode>>
```

Notes: 
The `setTernaryDarkMode` from the `useLocalStorage` uses the same technique and thus has the same signature.

lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -9,56 +9,63 @@ import { useUpdateEffect } from '../useUpdateEffect'
...
   }, [ternaryDarkMode, isDarkOS])
 
+  function toggleTernaryDarkMode() {
+    const toggleDict: Record<TernaryDarkMode, TernaryDarkMode> = {
+      light: 'system',
+      system: 'dark',
+      dark: 'light',
+    }
+    setTernaryDarkMode((prevMode: TernaryDarkMode) => toggleDict[prevMode])
+  }
```
juliencrn, line 54: Great refactoring 👍

Type not needed, `setTernaryDarkMode` already know it (hover the function in your IDE to see it) and `toggleDick` verifies it.
```js
setTernaryDarkMode(prevMode => toggleDict[prevMode])
```

lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -9,56 +9,63 @@ import { useUpdateEffect } from '../useUpdateEffect'
...
+
+  function dispatchTernaryDarkMode(newTernaryDarkMode?: TernaryDarkMode) {
+    if (newTernaryDarkMode) {
+      setTernaryDarkMode(newTernaryDarkMode)
+    } else {
+      toggleTernaryDarkMode()
+    }
+  }
```
juliencrn, line 62: I think it's better to have two separates methods that do only one thing each. The `newTernaryDarkMode` should be a required parameter (and remove the `if`).

But if we use `Dispatch<SetStateAction<T>>` type, we can delete this function and directly return the `setTernaryDarkMode`.

ps: in this kind of function, you may name the parameter just `mode`, there isn't name collision possible in the function scope, the function name tell us about the context and the parameter type too. The result will be a bit easier to read.

lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -9,56 +9,63 @@ import { useUpdateEffect } from '../useUpdateEffect'
...
-    toggleTernaryDarkMode: () => {
-      setTernaryDarkMode((currMode: TernaryDarkMode) =>
-        currMode === TernaryDarkMode.Dark
-          ? TernaryDarkMode.Light
-          : currMode === TernaryDarkMode.Light
-          ? TernaryDarkMode.System
-          : TernaryDarkMode.Dark,
-      )
-    },
+    dispatchTernaryDarkMode,
```
juliencrn, line 67: if we use Dispatch<SetStateAction<T>> type, you eventually may rename it into `setTernaryDarkMode` instead of `dispatchTernaryDarkMode` to avoid this:

```js
dispatchTernaryDarkMode: setTernaryDarkMode,
```

juliencrn (changes requested):
Hi @Pablion,

Cool, great work! 

There are some review comments, but only little changes, we are close to the release :)

You're welcome and thank you too, it's win-win 🤝

lib/src/useTernaryDarkMode/useTernaryDarkMode.ts
PabloLION: Nice call! I also considered this design was arguable.
And thanks for the naming advice.

PabloLION:
Hi @juliencrn . Obviously I did a wrong design and thank you for pointing it out.
I fixed all problems and updated the `.mdx`. Please check again.

juliencrn (approved):
Perfect, will pack it. Thank you, @PabloLION 🎉

juliencrn:
Released in the `v2.3.0`.

PabloLION:
🙂Thank you! It's my pleasure to collaborate with you!

juliencrn:
Haha, shit 😅  (np), and thanks ;)

You can submit another PR to improve the existing code if you want, but keep in mind to not introduce breaking changes for the current users (not the case here for the localstorage key), suit yourself! :)

PabloLION:
Agreed to your "no breaking changes", but I caused this in-perfection... and I wanted to fix it. 
Due to its relatively early, I think the change can happen. 
Reason:
1. The later we fix it, the worse it will be. 
2. Also other codes use this `ternaryDarkMode` will pollute the code. I want to change the key to `usehooks-ternary-dark-mode`. This will affect most users for the implemented code only once.
3. I actually wrote this in my note

> When use public space like `window`, `document`, `localStorage`, the naming should contain some package-wise unique identifier. From [gist](https://gist.github.com/juliencrn/8424b2965793aea4889e045a29409a3c)

See the full note [here](https://note.pablolion.com/) (I'll update the link after deploying)

juliencrn:
Thanks for your research, you're right, we should update local storage keys to at least prefix the key with "usehooks-ts". Good catch!

PR #109, opened 2022-02-15
Fix issue with useLocalStorage

bogdanailincaipnt:
I was having an issue with `useLocalStorage` in my project, and this change is solving it.

The problem was that when I had 2 tabs opened, and I changed the localStorage value in one of the tabs, the other tab would not always update correctly.
I tried reproducing this bug in a codesandbox but wasn't able to. It's probably a very specific/complex scenario that has to happen. I am using [SWR](https://swr.vercel.app/) to fetch some data based on values from localStorage & other places, but there are also other details that would be too hard to reproduce in a codesandbox.

Even without managing to reproduce this issue separately, I think my proposed change is harmless and I hope it will be accepted, or maybe even improved. Perhaps you guys can better understand why my change would make a difference.
Also, after testing, I concluded that this is the commit that introduced the issue: https://github.com/juliencrn/usehooks-ts/commit/bc3be93ad6a579272d02887b443ec9ff84560cb6 .
Maybe other hooks have similar issues.

Thanks in advance!

### DIFF
--- a/lib/src/useLocalStorage/useLocalStorage.ts
+++ b/lib/src/useLocalStorage/useLocalStorage.ts
@@ -1,4 +1,4 @@
-import { Dispatch, SetStateAction, useEffect, useState } from 'react'
+import { Dispatch, SetStateAction, useEffect, useState, useCallback } from 'react'
 
 // See: https://usehooks-ts.com/react-hook/use-event-listener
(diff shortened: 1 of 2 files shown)


juliencrn:
Hi @bogdanailincaipnt, thanks for the feedback and the fix. 

It looks good, I will include it in the next release.

PR #113, opened 2022-02-16
Feature/use state with callback

SleLLl:
Often in projects, we have to do something when the state has changed, and it's cool when you can change this function dynamically


juliencrn:
Hi @SleLLI,

Thank you for your contribution, but unfortunately, I will reject it. 

Indeed, have a callback attached to an `useState` is a frequent use-case and could be implemented as a React Hook. But here, I found that it doesn't really simplify the task. 
For example: Using `current.func` instead of a classical seams more complicated, I really think that the classical way to do it in React in simpler to read for everyone, and too, more flexible.

I like your concise `handleClick` with currying, setter and callback, all-in-one. But I have to spend little more time to understand it in comparison with a simpler approach.
As devs, we spend 7 times more time to read code than write, code should be optimized for fast reading/understanding, in my opinion.

Anyway, I will add you to the contributors list, and thanks for your time :)

SleLLl:
Hi @juliencrn, thanks for your consideration.

PR #114, opened 2022-02-17
Update useWindowSize.mdx

anver:
typo


juliencrn (approved):
Indeed, good catch! Thank you @anver

PR #117, opened 2022-02-22
apply lastest ref pattern to useEventListener.ts

TunA-Kai:
With the old implementation, user have to use `useCallback` to store their `handler` if they don't want the event to be removed and re-added every time the component rerender. By saving the `handler` in the `savedHandler` we can remove `handler` in the dependency list of `useEffect`. Therefore user don't have to remember to memoize their function in the first place.


juliencrn (approved):
It's nice. Thank you, @TunA-Kai 👍

PR #135, opened 2022-04-09
feat/add support for useMediaQuery in older versions of Safari

brycedorn:
Via [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener):

> In older browsers MediaQueryList did not yet inherit from [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget), so this method was provided as an alias of [EventTarget.addEventListener()](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener). Use addEventListener() instead of addListener() if it is available in the browsers you need to support.

These browsers (Safari <14) error with `'t.addEventListener' is undefined`, this adds support for those browsers and adds a note about it in docs.

### DIFF
--- a/lib/src/useMediaQuery/useMediaQuery.ts
+++ b/lib/src/useMediaQuery/useMediaQuery.ts
@@ -22,10 +22,18 @@ function useMediaQuery(query: string): boolean {
     handleChange()
 
     // Listen matchMedia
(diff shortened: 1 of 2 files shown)


juliencrn:
Hi @brycedorn, good catch, thank you 👍

PR #138, opened 2022-04-13
Added throttling and visual viewport support to useEventListener hook

SupraSmooth:
Hey 👋 
I really like this repo and wanted to contribute some code 😃 
During development I needed the `useEventListener` Hook but slightly tuned.

In this pull request, i've added optional parameters to throttle the fired handler and to use the Window.visualViewport as the fallback target for the event.

Those changes are fully optional and non-breaking. I think some other users can benefit from this too 💯 
I also added a jest test for the throttling option and updated the documentation.

Greetings from Germany!

### DIFF
--- a/lib/src/useEventListener/useEventListener.test.ts
+++ b/lib/src/useEventListener/useEventListener.test.ts
@@ -117,4 +117,29 @@ describe('useEventListener()', () => {
     expect(clickHandler).toHaveBeenCalledWith(expect.any(MouseEvent))
     expect(keydownHandler).toHaveBeenCalledWith(expect.any(KeyboardEvent))
   })
+
+  it('should throttle the event listener handler by the provided time', async () => {
+    const eventName = 'click'
+    const handler = jest.fn()
+
+    renderHook(() =>
+      useEventListener(eventName, handler, ref, { throttle: 20 }),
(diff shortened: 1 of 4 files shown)


juliencrn:
Hi @SupraSmooth, 

Thank you for this great contribution, I like this apport!

I did know the `visualViewport`, but it could be useful 👍 

However, I think the throttle function (good idea!)  shouldn't be inside the `useEventListener` hook, but as another hook whose could be composed with `useEventListener` or another hook. 

Using a throttle function with an event is a good practice, but the feature could require a debounce function instead of a throttle function by example. 

Thus, it should be 3 different hooks, letting the developer the freedom to compose together some wanted hooks to build its feature.

You have been created a throttle function yet, (and you can find the [debounce function here](https://usehooks-ts.com/react-hook/use-debounce) for inspiration), would you split the code by:
- removing just the throttle feature from this current PR
- creating another PR with the brand new `uesThrottle` hook (`npm run plop` to generate a new hook)

Thanks again from Spain!

SupraSmooth:
This feature needs some work to fit the general purpose of this repo.
If i find some time i'll create a `useThrottledEventListener` to handle function calls on events :)

PR #139, opened 2022-04-18
Fixed useFetch hook bug

daiky218:
When I was learning how to write about useFetch, I found that when the url changes, it will become the loaded state and will no longer change.(Data will not change)
This is because when the component is updated, cancelRequest will be set true. And will not be reset to false.I think this is bug.


juliencrn (approved):
Hi @daiky218, indeed, good catch! 👍  Thanks

PR #140, opened 2022-04-27
fix: useEventListener hooks to window when ref.current is null

zaggino:
There is a problem when we try to attach an event listener to an element that is conditionally rendered:
```
const Sample = () => {
  const btnRef = useRef(null);
  const onClick = useCallback(() => { alert('clicked btn'); }, []);
  useEventListener('click', onClick, btnRef);
  return (
    Math.random()
    ? <button ref={myRef}>click the button</button>
    : null
  )
}
```

In case button is not rendered, `element?.current` is null and it attaches the click handler to the window object instead, which is obviously not something that anybody wanted.


zaggino:
If this is a feature... then it's very hidden (and for my team unexpected) one... at least some warning to devs on https://usehooks-ts.com/react-hook/use-event-listener to not use the hook on conditionally rendered elements would be appropriate.

Issue #143, opened 2022-04-29
useEffectOnce may render two times with React.StrictMode


3GOMESz:
https://stackoverflow.com/questions/48846289/why-is-my-react-component-is-rendering-twice

3GOMESz:
Maybe a solution:

```
/* eslint-disable @typescript-eslint/no-unused-vars */
import {EffectCallback, useEffect, useRef} from 'react';

function useEffectOnce(effect: EffectCallback) {
  const destroyFunc = useRef<void | any>();
  const calledOnce = useRef(false);
  const renderAfterCalled = useRef(false);

  if (calledOnce.current) {
    renderAfterCalled.current = true;
  }

  useEffect(() => {
    if (calledOnce.current) {
      return;
    }

    calledOnce.current = true;
    destroyFunc.current = effect();

    return () => {
      if (!renderAfterCalled.current) {
        return;
      }

      if (destroyFunc.current) {
        destroyFunc.current();
      }
    };
  }, []);
}

export default useEffectOnce;

```

juliencrn:
Hi @3GOMESz, indeed yes, good catch! Thanks!

Would you submit a pull request with this change? If not, don't worry, I could do it later :)

3GOMESz:
Here it is : )

PR #152, opened 2022-05-15
Reverting useEffectOnce hook change

jherr:
The previous change was trying to circumvent checks done in React strict-mode. React strict-mode double mounts components to ensure that they mount and unmount cleanly. Effects should be written in such a way as they cleanup properly.

The code in the previous change also didn't work. The cleanup function is never called, either in strict mode, or in non-strict (i.e. production) mode.


juliencrn:
Hi @jherr, thanks for this PR. I've maybe accepted the #143 too fast, you're right 👍

PR #153, opened 2022-05-16
useLocalStorage: ignore unrelated 'storage' events (& fix perf issue)

bogdanailincaipnt:
Do not call `setStoredValue` for storage events whose `key` does not match current localStorage key (this also fixes some perf problems that can occur because of this - setStoredValue was being called too many times).

@juliencrn


juliencrn:
Nice improvement @bogdanailincaipnt 👍 Thank you

Issue #156, opened 2022-05-23
Improve useCountdown


PabloLION:
Hi again @juliencrn, I just moved back to react. I'm thinking on improving the `useCountdown` hook. But there would be breaking changes so I want to firstly have a small discussion with you.

## Breaking changes

In the [example usage](https://usehooks-ts.com/react-hook/use-countdown)

```typescript
  const [count, { start, stop, reset }] = useCountdown({
    seconds: 60,
    interval: 500,
    isIncrement: false,
  })
```

1. (important) Rename `seconds` to `countNumber` or `startingCount`, `countStart`, etc. Since it is not the real seconds.
2. (trivial) Change `start`, `stop`, `reset` to `startCountdown`, `stopCountdown`, `resetCountdown` respectively. The default one would be unclear and make the code harder to maintain.
3. (trivial) Change `interval` to `intervalMs` or `intervalMilisec` specifying its unit.

## Non-breaking changes
1. Add a `countLimit`, or `limit`, `countdownLimit`, `countEnd`, etc. to the setting.

## Outcome & alternative

The final outcome of this proposal would look like this:

```typescript
  const [count, { startCountdown, stopCountdown, resetCountdown }] = useCountdown({
    countStart: 60,
    intervalMs: 1000, // add 1000 by default
    countStop: 0, // 0 by default
    isIncrement: false,
  })
```

Alternative is to make a new `useIntervalTimer` hook to do the same job but with new name.

Thanks for reading. What's your opinion?

PabloLION:
Thanks for the confirmation. Working on it now.

> For the breaking change, it's ok, we will do minor update with a note in the changelog file 😎

On a second thought , maybe we can firstly keep the old functions and `console.warn()` about the deprecation when user uses the old ones, and then remove the old names in the next big version of usehooks-ts. What's your opinion?
Also, I'm thinking about using [@microsoft/tsdoc](https://www.npmjs.com/package/@microsoft/tsdoc) to add a `@deprecated` to make it clearer in VSCode, but there seem to be two problems:

1. I don't know where to add the TSDoc in this monorepo
2. Not sure if we can deprecate without changing the calling pattern.

I think you wouldn't mind me to ask the original creator @hexp1989 for his opinion on this too?

juliencrn:
Slowly depreciate the old interface is really a good idea 👍  

I don't know if a package is needed to depreciate a function, adding a `@depreciated` JSDoc comment  and the `console.warn()` could be enough.

[...]

PR #159, opened 2022-05-25
Docs/readme contribute node version

PabloLION:
Changing node version from `node@^16` to `node@16.x`, to specify that `node@{17,18}` is not compatible.
Also I could not commit nor push without `--no verify`, so I added another `### Tooling` part in the second commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,18 @@ npm run dev:site
 npm run test
 ```
 
+### Tooling
+
+Git hooks with husky will lint the code before commit and run tests before push. So before pushing, running the following command at least once is always needed.
+
+```bash
+# Launch the documentation website
+npm run bootstrap
+
+# before push
+npm run dev:site
+```
+
 ### How is structured a hook ?
 
 The hook itself and its unit tests are in the `/lib/src/` folder. It's the strict
--- a/README.md
+++ b/README.md
@@ -128,10 +128,11 @@ cd usehooks-ts
 npm run bootstrap
 ```
 
-### Create a new hook
+### Create or update a new hook
 
 ```bash
-# This command will regenerate all the hook boilerplate
+# This command generates boilerplate for new hooks.
+# Skip if updating an existed hook.
 npm run plop
 
 # Then develop the hook (aka test:watch)
@@ -143,18 +144,6 @@ npm run dev:site
 
 # Before commit: exec types-checking, linters and tests
 npm run test
-```
-
-### Tooling
-
-Git hooks with husky will lint the code before commit and run tests before push. So before pushing, running the following command at least once is always needed.
-
-```bash
-# Launch the documentation website
-npm run bootstrap
-
-# before push
-npm run dev:site
 ```
 
 ### How is structured a hook ?
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ If you need to make any other substantial changes, then follow the project setup
 This project use npm `Lerna` and `npm@8` to manage the different packages.
 Before starting, make sure you have the good system dependencies:
 
-- `node@^16`
+- `node@16.x`
 - `npm@^8`
 
 **Note**: To easily switch node version, consider Node Version Manager (nvm).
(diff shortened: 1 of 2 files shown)


PabloLION:
reason for first commit:
> It seems that node@18 was the problem here. After having a new git clone, node@18 still gives the same error, and with node@16 everything worked fine. Also I tested with node@17 (not working).


from: https://github.com/juliencrn/usehooks-ts/issues/156#issuecomment-1136567737

README.md
```diff
@@ -145,6 +145,18 @@ npm run dev:site
 npm run test
 ```
 
+### Tooling
+
+Git hooks with husky will lint the code before commit and run tests before push. So before pushing, running the following command at least once is always needed.
```
juliencrn, line 150: Yes, indeed!

README.md
```diff
@@ -145,6 +145,18 @@ npm run dev:site
...
+
+```bash
+# Launch the documentation website
+npm run bootstrap
+
+# before push
+npm run dev:site
+```
```
juliencrn, line 158: Hm, I'm curious about this. Basically, `npm run bootstrap` installs all the deps in the monorepo and should be executed at first, just after the `git clone`, but no after that. And `npm run dev:site` is launching the dev server, like a `npm start` or `npm run dev` with respectively CRA or Next.js app.
I don't understand why theses cmds help you to pass git hook requirements 🧐

Instead, consider these scripts: 

```sh
# build usehook-ts library (exec types check internally)
npm run build:lib

# Exec both all linter, types check and unit-tests
npm run test
```

Normally, these two commands will catch major codebase errors and then, allow a `git push`.

README.md
```diff
@@ -110,7 +110,7 @@ If you need to make any other substantial changes, then follow the project setup
 This project use npm `Lerna` and `npm@8` to manage the different packages.
 Before starting, make sure you have the good system dependencies:
 
-- `node@^16`
+- `node@16.x`
```
juliencrn, line 113: Ok, good catch, thanks 👍

juliencrn (changes requested):
Good idea to update the doc

README.md
PabloLION: I'll try to delete the repo and reinstall and see if it works on my side. Sorry for the delay.

PabloLION: here's the result of my tests:

1. husky pre-commit only works after `npm run bootstrap` so small changes can be made without initialization.
2. when a hook doesn't import from `'usehooks-ts'` the lint-staged won't need `build:lib`
3. publishing a branch will trigger `pre-push` and `npm run test` so the `npm run build:lib` is needed.

To sum up: after any code change with `npm run bootstrap`, the `build:lib` command is needed and the `npm run test` is not that much needed (if we use CLI to push, or haven't `npm run bootstrap`.) 

I'm thinking to merge this part to `### Create a new hook` or rename it to `### Update a existed hook`. What's your preference?

juliencrn: Your tests are ok, but I disagree with the shape.
We should recommend script for that it is intended to, I mean, `npm run bootstrap` isn't made for passing git-hooks and so, shouldn't describe in the doc as a requirement step to push. Instead, It should be run once at first to initiate the project locally. (We need to follow this rule for others script tough).

So, yes, you could create a `### Update an existing hook` block. 

```sh
# Install the whole workspace
npm run bootstrap

# Compile the library
npm run build:lib

# Exec Typescript, linter and unit tests
npm run test
```

These scripts aren't made to pass the git-hooks, the git-hooks are made to verify that the build steps are respected 🤣

Otherwise, it's a good plus to the doc, thanks 😊

PabloLION:
Hi, @juliencrn. After seeing you said
> These scripts aren't made to pass the git-hooks, the git-hooks are made to verify that the build steps are respected 🤣

I had a new idea: Maybe we can just add the `npm run build:lib` to the `bootstrap` command? This would save explaining all these stuff and seems "more correct" to me. With this setup we can push both before / after `npm run bootstrap`. (in comparison, in the current version after bootstrap and code change, lint/test for commit/push was not available)

juliencrn:
That make sense 👍 Thanks

PR #160, opened 2022-05-26
Use countdown/improve interface and enhance

PabloLION:
First part of issue #156. 
Adds a new version and keeping the deprecating old version.
With more test, and new demo & docs.

### DIFF
--- a/lib/src/useCountdown/useCountdown.test.ts
+++ b/lib/src/useCountdown/useCountdown.test.ts
@@ -20,10 +20,10 @@ describe('deprecated useCountdown()', () => {
     console.warn = jest.fn()
     renderHook(() => {
       useCountdown({ seconds: 60, interval: 500, isIncrement: false })
-      // expect(console.warn.mock.calls[0][0]).toBe(DEPRECATED_WARN)
-      // HELP! I cant type this `.mock`
     })
     expect(console.warn).not.toHaveBeenCalledWith('some random string')
+    // expect(console.warn.mock.calls[0][0]).toBe(DEPRECATED_WARN)
+    // HELP! I cant type this `.mock`
     expect(console.warn).toHaveBeenCalledWith(DEPRECATED_WARN)
   })
 
--- a/lib/src/useCountdown/useCountdown.test.ts
+++ b/lib/src/useCountdown/useCountdown.test.ts
@@ -92,15 +92,45 @@ describe('deprecated useCountdown()', () => {
 })
 
 describe('useCountdown()', () => {
-  test('should return callable functions', () => {
-    const { result } = renderHook(() =>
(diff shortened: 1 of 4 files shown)


lib/src/useCountdown/useCountdown.test.ts
```diff
@@ -1,10 +1,32 @@
...
+    return
+  }
+  test('should show deprecate warning', () => {
+    console.warn = jest.fn()
+    renderHook(() => {
+      useCountdown({ seconds: 60, interval: 500, isIncrement: false })
+    })
+    expect(console.warn).not.toHaveBeenCalledWith('some random string')
+    // expect(console.warn.mock.calls[0][0]).toBe(DEPRECATED_WARN)
+    // HELP! I cant type this `.mock`
```
PabloLION, line 26: I'd be appreciated if someone can tell me how to type the .mock because I didn't see answers by my search. Might ask later on Stack Overflow.

juliencrn:
Hi @PabloLION,

Great work 👏 I've fixed some stuff, but it's cool, thanks!

PabloLION:
Thanks. Please notify after creating the usehooks-ts@3 and I'll create another PR.

lib/src/useCountdown/useCountdown.ts
```diff
@@ -95,8 +91,7 @@ function useCountdown(
   // default values
   intervalMs = intervalMs ?? 1000
   isIncrement = isIncrement ?? false
-  countStop = countStop ?? 0
-  isIncrement = isIncrement ?? false
+  countStop = (countStop ?? 0) as number
```
PabloLION, line 94: Umm.. `countStop` is not actually a `number` (this got me for a while) because in the old version there was no such thing. So later on, in line 129 (old version) I had to skip calling `stopCountdown()` for the old version.

PabloLION: On a second look, my syntax seems to be wrong because this would never assign `undefined` to `countStop` yet the type hint of VSCode deduced `countStop: undefined | number` because later `countStop` was used in another scope.  Maybe we keep it your way, changing the behaviour of the hook and the countdown would stop at 0.

juliencrn: Good catch, I've tried to make this variable a number, but it can't. In the hook option, we indeed have `countStop: undefined | number`. I forgot to remove it to come back to `countStop = countStop ?? 0`. I will fix it, thanks!

PabloLION: Seems you misunderstood me but what you did was correct. The `countStop: undefined | number` would use the number `0` as a default value. What I was trying to do was to NOT add a `countStop` for the old hook since they might be designed not to stop. It seems that we should pass a `countStop = -Infinity` when `isDeprecated`.

PR #181, opened 2022-07-13
fix: strict equality in useLocalStorage

danbiilee:
Hi.
I changed equality to strict equality operator. I think it was typing mistake :)


juliencrn:
Hi, good catch 👍 Thanks

PR #183, opened 2022-07-18
Export interface from hooks

Lop3sPedro:
The objective is to export the project interfaces for reuse.


juliencrn:
Thanks for this proposition, but it's not wanted

PR #195, opened 2022-08-26
Fix useIntersectionObserver deps

Guesswhoitis:
useEffect has deps elementRef but elementRef never changes it is the .current we are interested in.

### DIFF
--- a/lib/src/useIntersectionObserver/useIntersectionObserver.ts
+++ b/lib/src/useIntersectionObserver/useIntersectionObserver.ts
@@ -35,7 +35,7 @@ function useIntersectionObserver(
     return () => observer.disconnect()
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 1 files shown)


gregkohn:
+1, would really love to see this merged -- just got bitten by it

hailwood:
Even then this won't actually fix it as the current value of a ref changing won't get picked up as a dependency change.

Note that [the blog post mentioned in the hook docs](https://medium.com/the-non-traditional-developer/how-to-use-an-intersectionobserver-in-a-react-hook-9fb061ac6cb5) has switched to using useState for this exact reason
image

juliencrn:
Hi guys, thanks for all. I'll merge it now, and plan to add the second version with the Dispatch function as a ref in the v3 because it introduces a breaking change.

juliencrn:
I can't merge it from here, the repo structure changed since. I'll replicate and mention it.

PR #197, opened 2022-08-29
Update useScript.ts

KATT:
- Make it so we get initial load status on first mount
- Add `cachedScriptStatuses`-record for short-circuting & avoiding DOM-lookup

~Potentially, we could have a global `Record<string, string|undefined>` defined rather than relying on `querySelector` to make it faster.~

### DIFF
--- a/lib/src/useScript/useScript.ts
+++ b/lib/src/useScript/useScript.ts
@@ -21,7 +21,6 @@ function getScriptNode(src: string) {
 
 /**
  * Load a script and return its status. If the script is already loaded, it will return it's load status immediately.
- * Modified to have an initial load status.
  */
 export function useScript(src: string): UseScriptStatus {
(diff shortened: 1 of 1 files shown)


lib/src/useScript/useScript.ts
```diff
@@ -1,70 +1,102 @@
...
+    }
+
+    return cachedScriptStatuses[src] ?? 'loading';
+  });
 
   useEffect(
     () => {
-      if (!src) {
-        setStatus('idle')
-        return
```
KATT, line 13: Deleted this, it felt weird to have an implicit `''` as an acceptable value, if anything I'd say we should do `string | null`

juliencrn:
Hi @KATT, 

Thanks for this interesting PR, unfortunately the repository structure have been changed since you've opened your PR.
I've created a new one using your code. I'll add your PR ID and your name in the commit message and in the release note.

Please, could you review this change?

Issue #203, opened 2022-09-02
[Suggestion] useScript


curtvict:
I'm curious if the community would find this change to the call signature of `useScript` at all useful:

```ts
function useScript(src: string, shouldLoad: boolean = true, removeOnUnmount: boolean = false): Status { ... }
```

`shouldLoad` could be used for something, like, an environment or privacy flag:

```ts
useScript('//some-script.js', isProduction) // or
useScript('//some-script.js', !isTest) // or
useScript('//some-script.js', user.isAnonymous)
```

`removeOnUnmount` could be used to remove the created script when the component calling `useScript` unmounts. Examples could be "one-shot" third party scripts that someone might just want to be present for a specific page of the app. Something like a HubSpot form that serves a specific use case, like Signing Up for an account.

If these sound like good ideas I'm happy to open a PR making the aforementioned changes.

juliencrn:
Hi @curtvict, 
Yeah, it's a cool idea, I'll be happy to review your PR. 

What about a signature like this ?
```ts
interface UseScriptOptions {
  shouldLoad?: boolean
  removeOnUnmount?: boolean
}

useScript(src: string, options?: UseScriptOptions): Status { ... } 
```

PR #220, opened 2022-09-30
feat: update useScript (#203, #197)

juliencrn:
This PR improves the `useScript` hook, inspired by:
- Issue #203 suggesting adding `shouldLoad` and `removeOnUnmount` options
- PR #197 adding in-memory caching

@KATT, @curtvict, I'd be happy to have your review :)

Notes: 
- It introduces breaking changes and should be added to the v3 branch
- Adding unit tests will be very appreciated

### DIFF
--- a/src/useScript/useScript.demo.tsx
+++ b/src/useScript/useScript.demo.tsx
@@ -24,7 +24,7 @@ export default function Component() {
     <div>
       <p>{`Current status: ${status}`}</p>
 
-      {status === 'ok' && <p>You can use the script here.</p>}
+      {status === 'ready' && <p>You can use the script here.</p>}
     </div>
   )
 }
--- a/src/useScript/useScript.demo.tsx
+++ b/src/useScript/useScript.demo.tsx
@@ -8,7 +8,9 @@ declare const jQuery: any
 
 export default function Component() {
   // Load the script asynchronously
-  const status = useScript(`https://code.jquery.com/jquery-3.5.1.min.js`)
+  const status = useScript(`https://code.jquery.com/jquery-3.5.1.min.js`, {
+    removeOnUnmount: false,
+  })
 
   useEffect(() => {
     if (typeof jQuery !== 'undefined') {
@@ -22,7 +24,7 @@ export default function Component() {
     <div>
       <p>{`Current status: ${status}`}</p>
 
-      {status === 'ready' && <p>You can use the script here.</p>}
+      {status === 'ok' && <p>You can use the script here.</p>}
     </div>
   )
 }
--- a/src/useScript/useScript.ts
+++ b/src/useScript/useScript.ts
@@ -1,14 +1,9 @@
 import { useEffect, useState } from 'react'
 
-export type UseScriptStatus = 'idle' | 'loading' | 'ok' | 'error'
+export type UseScriptStatus = 'idle' | 'loading' | 'ready' | 'error'
 export interface UseScriptOptions {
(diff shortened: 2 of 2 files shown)


src/useScript/useScript.ts
```diff
@@ -1,67 +1,115 @@
...
         }
 
-        script.addEventListener('load', setAttributeFromEvent)
-        script.addEventListener('error', setAttributeFromEvent)
+        scriptNode.addEventListener('load', setAttributeFromEvent)
+        scriptNode.addEventListener('error', setAttributeFromEvent)
       } else {
         // Grab existing script status from attribute and set to state.
-        setStatus(script.getAttribute('data-status') as Status)
+        setStatus(script.status ?? cachedScriptStatuses[src] ?? 'loading')
```
curtvict, line 88:
```suggestion
        setStatus(script.status ?? cachedScriptStatus ?? 'loading')
```

src/useScript/useScript.ts
```diff
@@ -1,67 +1,115 @@
...
-      script.addEventListener('error', setStateFromEvent)
+      scriptNode.addEventListener('load', setStateFromEvent)
+      scriptNode.addEventListener('error', setStateFromEvent)
 
       // Remove event listeners on cleanup
       return () => {
-        if (script) {
-          script.removeEventListener('load', setStateFromEvent)
-          script.removeEventListener('error', setStateFromEvent)
+        if (scriptNode && removeOnUnmount) {
```
curtvict, line 106: My original intent was to have `removeOnUnmount` actually remove the `scriptNode` from the DOM. I can't think of a use case where we wouldn't want to `removeEventListener`.

https://github.com/juliencrn/usehooks-ts/pull/223/files#diff-6ba736bc4c679006c0e6aac2fc18386cb2b6e5a371ff12712ed03fc03c7176f1R68-R70

src/useScript/useScript.ts
```diff
@@ -1,67 +1,115 @@
...
+export type UseScriptStatus = 'idle' | 'loading' | 'ok' | 'error'
+export interface UseScriptOptions {
+  shouldLoad?: boolean
+  removeOnUnmount?: boolean
+}
+
+const defaultOptions: UseScriptOptions = {
+  shouldLoad: true,
+  removeOnUnmount: true,
+}
```
curtvict, line 12: You wouldn't need these if you swap `shouldLoad` to `shouldPreventLoad` and `removeOnUnmount` actually removes the `scriptNode` itself.

curtvict (commented):
Just curious what you think of this.

src/useScript/useScript.ts
juliencrn: Indeed, my bad. Good catch! Thanks

curtvict: Yeah! No problem at all.

src/useScript/useScript.demo.tsx
```diff
@@ -22,7 +24,7 @@ export default function Component() {
     <div>
       <p>{`Current status: ${status}`}</p>
 
-      {status === 'ready' && <p>You can use the script here.</p>}
+      {status === 'ok' && <p>You can use the script here.</p>}
```
curtvict, line 27: This technically constitutes a breaking change, no? What was the reasoning for moving away from `'ready'`?

juliencrn:
Thanks for all @curtvict. This will be in the next release, probably the `2.8.0`.

PR #223, opened 2022-09-30
Add options to useScript

### LINKED ISSUE #203: [Suggestion] useScript

curtvict:
Closes #203

I ended up using `shouldPreventLoad` so that the default `undefined` could be falsey and it could go out in the next minor.

### DIFF
--- a/src/useScript/useScript.demo.tsx
+++ b/src/useScript/useScript.demo.tsx
@@ -1,3 +1,4 @@
+import React from 'react'
 import { useEffect } from 'react'
 
 import { useScript } from '..'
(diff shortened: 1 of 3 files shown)


juliencrn:
Have you seen this one #220 ?

juliencrn:
> I ended up using shouldPreventLoad so that the default undefined could be falsey and it could go out in the next minor.

Good idea 👍

PR #234, opened 2022-10-13
fix: rename `useBoolean`, `useCounter`, `useLockedBody` and `useSidebar` hooks ReturnType to Output

### LINKED ISSUE #233: `useBoolean`, `useCounter`, `useLockedBody` and `useSidebar` hooks uses `ReturnType` as output, witch is a reserved utility type in TS
This is causing a TS4023: Exported variable 'X' has or is using name 'ReturnType' from external module "/node_modules/usehooks-ts/dist/esm/useBoolean/useBoolean" but cannot be named.

valentinpolitov:
Closes #233


juliencrn (approved):
Good catch! Thanks

PR #235, opened 2022-10-13
fix: useEventListener passes options to removeEventListener

marnusw:
Options passed to `addEventListener` and `removeEventListener` must match for the listener to be removed. I found this issue when using an event in the capture phase.

### DIFF
--- a/src/useEventListener/useEventListener.ts
+++ b/src/useEventListener/useEventListener.ts
@@ -74,7 +74,7 @@ function useEventListener<
 
     // Remove event listener on cleanup
     return () => {
-      targetElement.removeEventListener(eventName, listener)
+      targetElement.removeEventListener(eventName, listener, options)
     }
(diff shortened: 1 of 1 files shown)


marnusw:
A different matter, but in the same file:

I'm wondering whether there is any reason to use the `useIsomorphicLayoutEffect` hook here:

```js
useIsomorphicLayoutEffect(() => {
  savedHandler.current = handler
}, [handler])
```

as opposed to just reassigning the ref on each render:

```js
savedHandler.current = handler
```

It doesn't seem like it would have any downsides, and it will simplify the hook and result in a very small performance improvement if the hook doesn't have to run and do the comparison on each render.

juliencrn (approved):
Good catch! Thanks

juliencrn:
I don't understand why, but this change breaks the unit tests

PR #237, opened 2022-10-16
Feature / Add UseOnline Hook

devrayat000:
A simple to track users' connectivity status.


TheHaff:
tip: might want to look into this https://www.freecodecamp.org/news/how-to-check-internet-connection-status-with-javascript/ 
the online event has problems as it doesn't really know if you have an internet connection.

PS. I'm often wrong

PR #242, opened 2022-10-24
useLocalStorage complies with useState

### LINKED ISSUE #204: useLocalStorage set function does not entirely comply with useState
the set function given by useState, guarantees that if it is called with a function argument, the first parameter will be the current value of the state. However, in useLocalStorage, this is not always true. in the case that a render does not happen between two set calls.

Example:

```javascript
const [n, setN] = useState(0);

useEffect(() => {
  setN(n => {
    console.log('state value:', n);
    return n + 1;
  });
  setN(n => {
    console.log('state value:', n);
    return n + 1;
  });
}, []);
```

This code will print out
```
0
1
```

useLocalStorage on the other hand:

```javascript
const [n, setN] = useLocalStorage('example-key', 0);

useEffect(() => {
  setN(n => {
    console.log('state value:', n);
    return n + 1;
  });
  setN(n => {
    console.log('state value:', n);
    return n + 1;
  });
}, [])
```
```
0
0
```

The problem stems from [line 55](https://github.com/juliencrn/usehooks-ts/blob/2dc20df615b32430b943349e8833b74a1fbc33dc/lib/src/useLocalStorage/useLocalStorage.ts#L55) where the stored value is used instead of reading the current value from localstorage. I copied the hook locally and changed it to `readValue()` and this fixed the issue.

Im sure theres something to be said here of performance, but i havent done any testing. possibly using a ref would be better suited for the task. Id be happy to whip up an implementation if this is something you're interested in!

Edit: off by one errors in code

valyrie97:
Fixes #204 

- add a test (that verified the issue)
- changed the parameter of the setState callback function to get a fresh value from localstorage.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "usehooks-ts",
-      "version": "2.7.0",
+      "version": "2.7.2",
       "license": "MIT",
       "workspaces": [
         "packages/eslint-config-custom"
--- a/src/useLocalStorage/useLocalStorage.test.ts
+++ b/src/useLocalStorage/useLocalStorage.test.ts
@@ -113,6 +113,20 @@ describe('useLocalStorage()', () => {
     expect(window.localStorage.getItem('count')).toEqual('3')
   })
 
+  test('Update the state with a callback function multiple times per render', () => {
+    const { result } = renderHook(() => useLocalStorage('count', 2))
+
+    act(() => {
+      const setState = result.current[1]
+      setState(prev => prev + 1)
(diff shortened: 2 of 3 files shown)


henriqemalheiros:
Ran into the same issue. `useSessionStorage` also has this problem.

wayneschuller:
This bug is causing me headaches! Would love to see the PR accepted...

juliencrn:
Thanks @valyrie97 that's perfect 👌
As the base branch is a bit outdated, I applied your change here #436 on both `useLocalStorage` and `useSessionStorage`

PR #245, opened 2022-11-07
feat: Improve useFetch

CaptainN:
- Allows refetch when url changes
- Allows url to be null to skip loading (ex: to defer action)
- Properly track mount (`inEffect`) instead of using useMemo 
   - Fixes edge case errors with component re-use 
      Example: If used in React Router - refetch will now happen when url or params change.
- Provide loading meta
- Extend API to allow parse method spec (text or json)
- Change return type to more natural tuple

These changes allow for additional hooks to be built on top, for example, here is a `useFetchLater` (or `useFetchPost`) hook, which returns the same data, with a submit handle:

```ts
import { useState } from "react";
import useFetch from "./useFetch";

const useFetchPost = <T>(url: string) => {
  const [postData, setPostData] = useState({
    url: undefined,
    values: "{}",
  });

  const fetchState = useFetch<T>(postData.url, {
    method: "post",
    credentials: "same-origin",
    body: postData.values,
  });

  return [
    function submit(values) {
      setPostData({
        url,
        values: JSON.stringify(values),
      });
    },
    ...fetchState,
  ] as const;
};

export default useFetchPost;
```

(That still has limitations that could be improved upon, but it wasn't possible with the old `useFetch` implementation.)

### DIFF
--- a/src/useFetch/useFetch.ts
+++ b/src/useFetch/useFetch.ts
@@ -1,89 +1,98 @@
-import { useEffect, useReducer, useRef } from 'react'
+import { useEffect, useReducer } from 'react'
 
-interface State<T> {
+interface useFetchState<T> {
   data?: T
   error?: Error
+  loading: boolean
 }
 
-type Cache<T> = { [url: string]: T }
-
-// discriminated union type
 type Action<T> =
   | { type: 'loading' }
   | { type: 'fetched'; payload: T }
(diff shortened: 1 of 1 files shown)


juliencrn:
Hi @CaptainN,
Great work! Thank you. 
Unfortunately, I'll close it because `useFetch` will be deprecated and dropped soon.

Issue #176, opened 2022-07-02
useSessionStorage causing React+Nextjs application crash. Error "Hydration failed because the initial UI does not match what was rendered on the server"


roker15:
My application was working fine. I was required to refresh pages without losing state. so i replaced `useState `with `useSessionStorage `wherever required. Now it throwing error
`Hydration failed because the initial UI does not match what was rendered on the server`

This [Link](https://nextjs.org/docs/messages/react-hydration-error) might help but not sure.

harrywebdev:
I reckon this is most likely because the `useState` hook doesn't use the provided `initialValue` here
https://github.com/juliencrn/usehooks-ts/blob/master/src/useSessionStorage/useSessionStorage.ts#L39

which is, what is expected on client on the first render. I think the line should read

```
const [storedValue, setStoredValue] = useState<T>(initialValue)
```

and if I understand the code correctly, then this comment

> // Pass initial state function to useState so logic is only executed once

is actually handled by the `useEffect(..., [])` here
https://github.com/juliencrn/usehooks-ts/blob/master/src/useSessionStorage/useSessionStorage.ts#L69

I'll see if I can create a PR - maybe in case @juliencrn or @createdbymahmood think that's a sound idea/fix?

juliencrn:
Should have been fixed by #435, will be in the next release. Thanks all!

PR #251, opened 2022-11-16
fix: useSessionStorage hydration mismatch

harrywebdev:
Initial value for `useState` hook should be identical to the one provided in SSR to prevent hydration mismatch error.

### DIFF
--- a/src/useSessionStorage/useSessionStorage.ts
+++ b/src/useSessionStorage/useSessionStorage.ts
@@ -35,8 +35,8 @@ function useSessionStorage<T>(key: string, initialValue: T): [T, SetValue<T>] {
   }, [initialValue, key])
 
   // State to store our value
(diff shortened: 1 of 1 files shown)


harrywebdev:
This should fix https://github.com/juliencrn/usehooks-ts/issues/176

harrywebdev:
@juliencrn maybe the same should be done for localStorage? Should I also add that in?

harrywebdev:
does anyone else manage this repo?

sverps:
Ran into the same issue you're having, indeed with the `useLocalStorage` hook. Fix looks fine, would be nice to get this rolled out.

juliencrn:
Hi @harrywebdev, thanks for your PR. The base branch was a bit outdated but I've taken your changes in #435 mentioning you

PR #257, opened 2022-12-04
Fix useLocalStorage API

pedrobslisboa:
The way `useLocalStorare` was built does not reflect useState behavior. 

You guys are passing the `getter` to the `setter` instead of getting the previous value, which does not reflect the useState API. It can produce a lot of bugs since the useState api works async and need to get the previous value on the setter, instead of directly the getter, to work correctly.

You can check out the problem running it here: https://codesandbox.io/s/uselocalstorage-sample-mistake-uxms2b

You can check out the solution working here: https://codesandbox.io/s/uselocalstorage-sample-working-right-4ls3ox?file=/src/App.tsx

Same issue:
https://github.com/uidotdev/usehooks/pull/156

### DIFF
--- a/src/useLocalStorage/useLocalStorage.ts
+++ b/src/useLocalStorage/useLocalStorage.ts
@@ -49,17 +49,26 @@ function useLocalStorage<T>(key: string, initialValue: T): [T, SetValue<T>] {
     }
 
     try {
-      // Allow value to be a function so we have the same API as useState
-      const newValue = value instanceof Function ? value(storedValue) : value
-
-      // Save to local storage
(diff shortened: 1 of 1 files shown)


pedrobslisboa:
@juliencrn

henriqemalheiros:
Ran into the same issue. There is an older PR trying to solve it too (https://github.com/juliencrn/usehooks-ts/pull/242). Yours is missing the `window.dispatchEvent` that notifies other `useLocalStorage`s that have the same key.

juliencrn:
Fixed by #436, thanks all

PR #262, opened 2022-12-11
minor fix

samay-rgb:


juliencrn:
Hi @samay-rgb,
Good catch but I'll close the PR since useSsr is depreciated and will be removed soon.
Thanks

PR #272, opened 2023-01-02
feat: add reviver support to useLocalStorage

BalzGuenat:
This adds the optional `reviver` parameter of `JSON.parse` to `useLocalStorage`.

If this sounds like a feature that should be added, let me know and I'll extend the MR with docu and tests.


juliencrn:
Hi @BalzGuenat,
Thank you for this PR, it was a great idea and implementation 👍
Unfortunately, I'll close it because the code has been changed since your work (making rebasing tricky), and the problem was fixed another way, you now can pass your own custom `serializer` and `deserializer`. 
Thanks again anyway!

PR #276, opened 2023-01-09
useSessionStorage improvements

sskirby:
This PR has 3 changes for the `useSessionStorage` hook:

1. It allows an initial value function to be used without provoking TS errors.
2. It eliminates changes to the hook state when another instance of the hook with a different storage key is updated.
3. It ensures that the value set to the hook is referentially equal to the value reported by the hook.

I love the library! Please let me know how I can improve the PR for your approval!

### DIFF
--- a/src/useSessionStorage/useSessionStorage.ts
+++ b/src/useSessionStorage/useSessionStorage.ts
@@ -16,21 +16,26 @@ declare global {
 
 type SetValue<T> = Dispatch<SetStateAction<T>>
 
-function useSessionStorage<T>(key: string, initialValue: T): [T, SetValue<T>] {
+function useSessionStorage<T>(
+  key: string,
+  initialValue: T | (() => T),
+): [T, SetValue<T>] {
   // Get from session storage then
   // parse stored json or return initialValue
   const readValue = useCallback((): T => {
+    const getInitialValue = () =>
+      initialValue instanceof Function ? initialValue() : initialValue
     // Prevent build error "window is undefined" but keep keep working
     if (typeof window === 'undefined') {
-      return initialValue
+      return getInitialValue()
     }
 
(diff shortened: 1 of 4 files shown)


src/useSessionStorage/useSessionStorage.ts
```diff
@@ -16,27 +16,34 @@ declare global {
...
-      return initialValue
+      return getInitialValue()
     }
   }, [initialValue, key])
 
   // State to store our value
   // Pass initial state function to useState so logic is only executed once
   const [storedValue, setStoredValue] = useState<T>(readValue)
+  // create a new empty object to keep track of the identity of the hook
+  const [source] = useState({})
```
sskirby, line 46: @juliencrn do you think it worthwhile to extract the creation of the empty object into a function so that it isn't created on each render?

src/useSessionStorage/useSessionStorage.ts
```diff
@@ -16,27 +16,34 @@ declare global {
...
-function useSessionStorage<T>(key: string, initialValue: T): [T, SetValue<T>] {
+function useSessionStorage<T>(
+  key: string,
+  initialValue: T | (() => T),
+): [T, SetValue<T>] {
   // Get from session storage then
   // parse stored json or return initialValue
   const readValue = useCallback((): T => {
+    const getInitialValue = () =>
+      initialValue instanceof Function ? initialValue() : initialValue
```
sskirby, line 27: @juliencrn or maybe memo-ize the initialValue so that we guarantee it only ever gets evaluated once?

juliencrn:
Hi @sskirby, thanks for the PR. I will unfortunately close it because most of the changes already have been addressed now, see #435, #436 and #437

PR #280, opened 2023-01-13
Fix user of useScreen not rerendering on screen resize

philipgher:
The component that uses the useScreen hook will not rerender in it's current state due to window.screen maintaining it's reference. Thus calling setScreen is not triggering a rerender. Fixed by creating a shallow clone of the new screen object.


juliencrn:
Hi @philipgher, 
Thanks for the PR, unfortunately, this PR is too old now, but I applied your change to #474 and will add you to the contributor list. The fix will be in the next release

PR #285, opened 2023-01-26
adds id as option to useScript hook

misidoro:
I am currently using useScript to download zendesk script but an id is required to make it work. It's working on the project as I copied the whole hook and added the id but I think it would be a good addiction.

Currently using it as
```
const status = useScript("https://static.zdassets.com/ekr/snippet.js?key=8f60d464-1c48-48a6-8d14-25f47c389e59", {
    id: "ze-snippet"
});
```

In order to get the project running I had to install a bunch of eslint dependencies. Not sure if I am doing something wrong.

I was not able to get the docs running, might be due to some setup missing.


juliencrn:
Hey @misidoro, 
Great add! Thank you. I moved your work there (#525) because this PR is too old and the code moved since.  
I added you to the contributor list.

PR #260, opened 2022-12-07
Fix minor typo in documentation

michaelmior:

PR #287, opened 2023-01-27
docs: fix hook name typo

hugohutri:
A typo fix


juliencrn:
Hi @hugohutri, good catch but I've already merged this change here 135920a8 from #260 by @michaelmior. Thanks! 👍

PR #293, opened 2023-02-12
refactor(useIsMounted): :art: use useMemo instead of useCallback

lawlesx:
Refactored `useIsMounted` to use `useMemo` instead of `useCallback`. This helps in not calling the `isMounted` as `isMounted()` i.e `()` is not required.

This shortens the code and directly use the value of the hook instead of calling it and then using the value.

### DIFF
--- a/src/useIsMounted/useIsMounted.demo.tsx
+++ b/src/useIsMounted/useIsMounted.demo.tsx
@@ -11,7 +11,7 @@ function Child() {
   // simulate an api call and update state
   useEffect(() => {
     void delay(3000).then(() => {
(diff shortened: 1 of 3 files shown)


ChanhyukPark-Tech (approved):
godo

juliencrn:
Hi @lawlesx,
Thanks for the PR. I agree with you that `isMounted` is the nicer API than `isMounted`. 
Unfortunately, it seems to be not working as expected. 
I tried both manual testing using the `.demo.tsx` and unit tests, but no one worked on my side.

PR #297, opened 2023-02-21
feat: create useOS hook

Moyo75:
- added useOS hook 
- fixed some grammar mistakes


Moyo75:
I added a new hook called useOS and fixed some writing mistakes. Though I was able to build the library, requests for the individual hooks pages failed.

PR #317, opened 2023-05-12
Delete item from cachedScriptStatuses when script node is deleted

gugu:
Currently, the node is deleted on unmount, but next time the same element mounts the script is not included


juliencrn:
Thanks for the PR, it does fix the bug, but it has been fixed in #354

PR #326, opened 2023-05-30
Use callback for countdown reset

gromchen:
Otherwise, it's a potential infinite loop in `useEffect`.


juliencrn:
Hi @gromchen,

Thanks for that PR, since the code is too old, I re-created it there (#526).
I added you to the contributor list.

PR #334, opened 2023-06-08
Update useInterval.ts

jcsumlin:
If all we care about is making sure the delay is set to a number wouldn't this simplify that logic?


juliencrn (approved):
Good 👍 Thanks

PR #338, opened 2023-06-22
chore: add export on usebooleanoutput interface

Joshyahweh:
Add export on usebooleanoutput interface so that it can be used globally.


juliencrn:
Hi @Joshyahweh, you can do this instead: 

```ts
import { useBoolean } from 'usehooks-ts'

type UseBooleanOutput = ReturnType<typeof useBoolean>;
```

PR #339, opened 2023-06-26
Update useLocalStorage.ts - Include 'Key' in useEffect which updates …

jackdh:
…stored value.

I've found this was required should the key be dynamic, as otherwise it will continue to use the old key.


juliencrn:
Hi @jackdh, thanks for your contribution but I've unfortunately just merged the same PR #355

Issue #335, opened 2023-06-10
useSessionStorage returns undefined if the value is already set as a string (+ some suggestions)


jonlinkens:
`useSessionStorage` will return `undefined` if a string value already exists in session storage under the same key. This is because the hook always parses state as a JSON object with `JSON.parse`, which will error on strings and thus return `undefined`.

This can be avoided by using `JSON.stringify` on the string you initially set in session storage, but this feels a little too opinionated and cumbersome to do everywhere you have this case.

This would be useful for places in code where hooks can't be used to initially set a value in session storage - i.e. at app initialisation, and then accessing/setting it via the hook elsewhere.

### Steps to reproduce 
https://codesandbox.io/s/loving-taussig-m3qtkm?file=/src/App.tsx

As in the sandbox, set a string value in session storage beforehand, then try and access it via the hook. You'll see a `parsing error on ...` error gets printed to the console.

This behaviour can also be seen with this test case failing:
```ts
test('existing string state is in the returned state', () => {
  window.sessionStorage.setItem('key', 'value')

  const { result } = renderHook(() => useSessionStorage('key', 'value'))

  expect(result.current[0]).toBe('value')
})
```

giving the output

```shell
 Expected: "value"
 Received: undefined
```

### Further comments
As well as the above, `initialValue` is never set in session storage - which feels like expected behaviour to come from this hook.

Additionally, there is no equivalent to `useReadLocalStorage` where one can pass through only a session storage key and get the value (or return `undefined` if doesn't exist) like so:

```ts
const value = useReadSessionStorage('test-key')
```

### Proposed changes
- To fix the existing string returning as undefined bug, I see two options:
    - Use `JSON.stringify` on the `item` value inside the hook, although I think this might introduce some issues
    - Become less opinionated, by adding some an optional options object argument:
        - `parseAsJSON: boolean` - default to `true` to avoid breaking changes
        - `parser: (value: string | null) => T | undefined` - allow custom parsers to be used in place of the existing `parseJSON` inside the hook
        - An accompanying `serializer` function
- Set the `initialValue` in session storage, if one doesn't already exist
- Add a `useReadSessionStorage` hook with similar behaviour to `useReadLocalStorage`
- `initialValue` is a bit of a confusing name - right now it acts more as a `defaultValue`, so could rename to that (although with the `setItem` change above this would be fine as is)

***

I'd be more than happy to pick this up in a PR, can hopefully submit some time soon.

simbesh:
[...]

PR #340, opened 2023-06-26
Update useSessionStorage to support strings + add new options + add useReadSessionStorage

jonlinkens:
Fixes https://github.com/juliencrn/usehooks-ts/issues/335

Attempts to make `useSessionStorage` less opinionated in how session storage values are stored and parsed.

### Changes
- Add third `options` argument to `useSessionStorage`:
    - `parseAsJson: boolean` - default to true, pass as `false` when you don't want a value to be parsed using `JSON.parse`. This will fix the bug I was describing where a value has been previously set as a `string`
    - `parser: (value: string | null) => T` - a custom parser if you have stored a session storage value with a custom serializer function
    - `serializer: (value: T) => string` - a custom serializer to store values in a format different from the default behavior of `JSON.stringify`
- Renamed `initialValue` to `defaultValue` for readability
- Removed `undefined` value support - I can't imagine a case where this is needed over just not storing the key/value in session storage + removing this decreases complexity of the hook
    
- Add `useReadSessionStorage`, similar to `useReadLocalStorage`


jonlinkens:
Going to close this as it seems this library is no longer maintained

jonlinkens:
I think this can now be closed as addressed in https://github.com/juliencrn/usehooks-ts/pull/439?

juliencrn:
> I think this can now be closed as addressed in #439? I'll leave open so you can confirm - Thank you very much btw!

Yes indeed! Thank you 👍 I will add you to the contributor list

jonlinkens:
>  I, unfortunately, worked on it before seeing your one

No worries at all, just happy to see the bug fixed - thanks for your hard work on this library

erichartline:
Any chance that we could still bring in the `useReadSessionStorage` hook from this PR? We have the equivalent for localStorage but not sessionStorage.

PR #342, opened 2023-07-01
fix(Docs): external link to appear

ojj1123:
As a new official document was created in this PR https://github.com/juliencrn/usehooks-ts/pull/333, the element of moving to an external link in the document disappeared.

Check ➡ [use-intersection-observer](https://usehooks-ts.com/react-hook/use-intersection-observer)

## TO-BE

In addition, there was an issue where the element going to the external link was not visible on other custom hook document pages, so it was corrected.

### DIFF
--- a/apps/www/src/components/remote-mdx.tsx
+++ b/apps/www/src/components/remote-mdx.tsx
@@ -58,13 +58,15 @@ export const SmartLink = ({
   }
 
   return (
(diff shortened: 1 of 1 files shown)


juliencrn:
Oh good catch @ojj1123, your fix was good but I've already fixed this bug before seeing your PR. Thanks!

Issue #266, opened 2022-12-18
The `useFetch()` hook is hardcoded to work only with JSON


haroldo-ok:
If I try to use `useFetch()` on some API endpoint that returns plain text or an image, it returns an error because `useFetch()` is hardcoded to parse the response body as JSON.

juliencrn:
Hi, thanks for the bug report! 

I'll close the issue because `useFetch` will be deprecated and then dropped. 
Instead, I suggest you to use either:
- Your framework stuff (like RSC or getServerSideProps from NextJS for example)
- `use` from React
- Libraries like `react-query` or `useSWR`

DerGoogler:
@haroldo-ok @figurluk 

Use this: https://github.com/DerGoogler/MMRL/blob/master/Website/src/hooks/useFetch.ts

```js
const [mData, mError] = useFetch("url", { type: "text" })
```
> Default is `json`

PR #343, opened 2023-07-03
Add support on useFetch

lit26:
- Add loading state
- Support for other types of response data
- Fix https://github.com/juliencrn/usehooks-ts/issues/266

### DIFF
--- a/packages/usehooks-ts/src/useFetch/useFetch.ts
+++ b/packages/usehooks-ts/src/useFetch/useFetch.ts
@@ -3,6 +3,7 @@ import { useEffect, useReducer, useRef } from 'react'
 interface State<T> {
   data?: T
   error?: Error
+  isLoading: boolean
 }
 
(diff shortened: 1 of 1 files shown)


juliencrn:
Hi @lit26, 
Thanks for the PR but we'll depreciate this hook.

tli26:
> Hi @lit26, 
> 
> Thanks for the PR but we'll depreciate this hook.

Why this is closed?

juliencrn:
Because `useFetch` will be depreciated and dropped soon. 

I recommend you to use instead:
- your framework server-side stuff like RSC or getServerSideProps  for NextJS
- use the `use` hooks from react
- use `useQuery` or `useSWR` 

Our `useFetch` was a good example to learn how a react hook works, but it's an anti-pattern and not enterprise-grade.

tli26:
I see. Thanks

PR #344, opened 2023-07-06
feat(useTernaryDarkMode): Add `defaultValue` parameter

RobHannay:
This parameter will allow the user to set a default value for the dark mode when there's no value in the local storage. 

Documentation is also updated to reflect these changes.

## Use case

We want to allow users to choose a dark mode option, but for it to default to light - not to follow the system by default. This is because we want to slowly enable it as opt-in whilst it's in beta, but keep the 'system' option for this toggle.


packages/usehooks-ts/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -26,36 +27,14 @@ interface UseTernaryDarkModeOutput {
...
-        setDarkMode(false)
-        break
-      case 'system':
-        setDarkMode(isDarkOS)
-        break
-      case 'dark':
-        setDarkMode(true)
-        break
-    }
-  }, [ternaryDarkMode, isDarkOS])
```
RobHannay, line 54: Have simplified this state-syncing logic as I don't believe it's necessary - it can just be derived from the above local storage & media query states.

packages/usehooks-ts/src/useTernaryDarkMode/useTernaryDarkMode.ts
```diff
@@ -26,36 +27,14 @@ interface UseTernaryDarkModeOutput {
  */
 export function useTernaryDarkMode(
   localStorageKey: string = LOCAL_STORAGE_KEY,
+  defaultValue: TernaryDarkMode = 'system',
```
RobHannay, line 30: I've resolve the coflicts here and put it as the second param to avoid introducing a breaking change.

However, this will also mean that to set a default value, people will have to specify a `localStorageKey`, even if they just want to use the default key.

I would be in favour of introducing object-params here so that it's possible to specify whichever optional params you want, like

```ts
useTernaryDarkMode({ defaultValue: 'light' });
```

which avoids having to re-specify the default local storage key just to set a default

```ts
useTernaryDarkMode(LOCAL_STORAGE_KEY, 'light')
```

We could do this without it being a breaking change with method overloading. What do you think @juliencrn?

juliencrn: Hi @RobHannay,
Thanks for the PR (and its recent update). 

Adding a `defaultValue` and making an object params is a great idea 👍 
It effectively introduces API change, so I've indeed used function overload to accept both the old and the new interface. The old one will be dropped in the v3

juliencrn:
@RobHannay thanks for all! I've done the same for `useDarkTheme` too and added some unit tests on top of that.

PR #350, opened 2023-07-15
Fix `useIsClient()` refrence invisible in useSsr.md

LarsArtmann:
+added `useIsClient()` to `Related hooks` like it is in [use-session-storage](https://usehooks-ts.com/react-hook/use-session-storage)

Find problem after: [use-ssr()](https://usehooks-ts.com/react-hook/use-ssr#:~:text=Otherwise%2C%20If%20you%20want%20to%20be%20notified%20when%20the%20value%20changes%20to%20react%20to%20it%2C%20you%20can%20use)


juliencrn:
Hi @LarsArtmann, you're right but I'll close it since `useSsr` has been depreciated and will be removed soon. Thank you anyway! 👍

Issue #314, opened 2023-05-02
useScript cleanup on removeOnUnmount


sebadom:
I'm using useScript in an nextjs app, to load an external javascript snippet that runs it's initialization logic on load, therefore whenever I navigate to another page I'm using `removeOnUnmount` in true, with the hope that any further navigation to any other page using this widget will re init everything. Now the thing is I've seen the code and while it disables the listeners and remove the node, it doesn't actually remove the cached key from `cachedScriptStatuses` therefor it does not recreate the node again. 

Is that a bug or is that by design?

Thank you

ShanSenanayake:
I have run into the same issue and is also wondering about the same thing. 
The cached script isn't remove as a key which means if we want to mount the component again it assumes the script is already loaded, which is not the case when `removeOnUnmount` is set to true. I would assume this is a bug.

juliencrn:
Hi, thanks for the feedback, it has been fixed in #354 and will be released in days

PR #354, opened 2023-07-20
fix: useScript failed to remove script from cache when passing `removeOnUnmount` prop

ShanSenanayake:
The `removeOnUnmount` prop didn't remove the script from the cache. This created a side effect where if you try and render the same component by mounting -> unmounting -> mounting again, the `useScript` hook behaves as the script is loaded when in reality is has remove the script from the tree but failed to remove it from the cache.

Issue: https://github.com/juliencrn/usehooks-ts/issues/314


scottwiw:
Any bump on this? Ran into this issue while trying to add a Hubspot calendar iframe and this definitely fixes it.

ShanSenanayake:
@scottwiw based on [this issue](https://github.com/juliencrn/usehooks-ts/issues/373), and the amount of PRs that has no activity I think this package is basically abandoned, I would suggest co locating the whole `useScript` and fixing it yourself.

scottwiw:
> @scottwiw based on [this issue](https://github.com/juliencrn/usehooks-ts/issues/373), and the amount of PRs that has no activity I think this package is basically abandoned, I would suggest co locating the whole `useScript` and fixing it yourself.

Thanks, I'll make do. Appreciate it!

juliencrn:
Thank you @ShanSenanayake 👍

PR #355, opened 2023-07-22
feat: now useLocalStorage will recalculate default value if key changes

amirking59:
if your key is dynamic 
now this hook is able to recalculate default value


juliencrn (approved):
Good, thanks!

PR #360, opened 2023-07-29
feat: create useCloseAlert hook

vsnthdev:
This PR adds a new hook called `useCloseAlert()` that will prevent the page from closing for a few seconds, when the app is sending API calls, or being busy saving user's changes.

It is useful in preventing accidentally immediate closing of apps that are in the process of saving something.

Popular websites like Gmail, Spotify, Twitter and Notion use this technique as well.

Thank you 😀


vsnthdev:
CLosing this PR as it looks like the create has abandoned  this package.

PR #366, opened 2023-08-10
Fix bug with useLocalStorage.ts hook

danielturus:
I am having an issue with `useLocalStorage` in my project and this PR tries to solve it.

Problem:

I have two instances of useLocalStorage hook (different keys).
One updates localStorage (via setter function) on event click on `document` object and the other, since it listens for the `local-storage` custom event gets also triggered and updates its state.

```ts
  useEventListener('local-storage', handleStorageChange)
 ```

In order to update the state of each instance of useLocalStorage hook we need to make sure that the received event key is  matching our instance's key.

Thanks!

### DIFF
--- a/packages/usehooks-ts/src/useLocalStorage/useLocalStorage.ts
+++ b/packages/usehooks-ts/src/useLocalStorage/useLocalStorage.ts
@@ -15,6 +15,9 @@ declare global {
 }
 
 type SetValue<T> = Dispatch<SetStateAction<T>>
+type LocalStorageEventDetail = {
+  key: string;
+}
 
 export function useLocalStorage<T>(
   key: string,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/usehooks-ts/src/useLocalStorage/useLocalStorage.ts [hunk 1]
```diff
@@ -62,7 +65,11 @@ export function useLocalStorage<T>(
       setStoredValue(newValue)
 
       // We dispatch a custom event so every useLocalStorage hook are notified
-      window.dispatchEvent(new Event('local-storage'))
+      window.dispatchEvent(new CustomEvent<LocalStorageEventDetail>('local-storage', {
+        detail: {
+          key
+        }
+      }))
```
fritzmonkey, line 72: I ran into this issue as well, and came up with this solution. This way the synthetic event matches the browser events and the handler logic doesn't need to change.
```suggestion
      const event = new Event('local-storage')
      event.key = key
      window.dispatchEvent(event)
```

Mind implementing a fix in the `useSessionStorage` hook as well? It has the same bug.

PR #374, opened 2023-08-29
fix: to not call handler when disconnected target's event occurs

hooriza:
Fixed a bug with `handler` being called when the event target is not connected to a document

I've created a sample code to check the bug
https://codesandbox.io/s/gracious-ardinghelli-cx6j84?file=/src/Dialog.tsx


juliencrn (approved):
Great, thank you 👍

PR #380, opened 2023-09-14
chore(useSet): add use set hook

gzhgutov:


dretsa:
Sorry for bringing this PR from the dead. I think it would be great to get it merged 💯 
I find it strange that there is `useMap` but not `useSet`

PR #383, opened 2023-09-23
Removed the need of commenting out useEffect deps

lisandro52:
A little QoL update, we don't really need to comment the dependency array warning, as it's bad practice if it's avoidable.


juliencrn (approved):
Well done, thanks

PR #385, opened 2023-09-28
Proposal: add `post` method to `useFetch`

iamwillnbcu:
# Proposal for Extending `useFetch` Hook to Support `POST` Method and Caching

## Introduction

The `useFetch` hook is a utility for making HTTP requests in a React application. Currently, it only supports fetching data (`GET` requests). This proposal outlines the need for extending the hook to include a `POST` method for submitting data.

## Objective

To extend the `useFetch` hook to include a `POST` method that allows the user to send data to a server endpoint.

## Main Features

- **Caching**: The hook uses a simple cache mechanism to avoid fetching the same data multiple times.
- **Post Request**: It exposes a `post` function that can be used to perform `POST` requests. This function also has various customization options.
- **Immediate State Update**: The hook provides an option to update the state immediately after a `POST` request, which can be particularly useful for optimistic updates.
- **Refetch after POST**: The hook can also refetch data from the original `GET` URL after a successful `POST` request, ensuring that the client state is synchronized with the server.
- **Custom Headers**: You can pass custom headers for both `GET` and `POST` requests.
- **Error Handling**: The state object returned by the hook contains an `error` field that will be populated if an error occurs during fetch.
- **Success and Failure Callbacks**: The `post` function accepts `onSuccess` and `onFailure` callbacks that will be executed based on the result of the `POST` request.

## Motivation

1. **Completeness**: Adding a `POST` method would make the `useFetch` hook more complete and versatile as a utility for HTTP operations.
2. **Ease of Use**: Developers using `useFetch` won't have to write custom `POST` logic every time they need to submit data.
3. **Consistency**: A single, consistent API for both fetching and posting data will improve developer experience.
4. **Code Reusability**: The `POST` method can leverage existing internal logic, reducing redundancy.

## Proposed Method Signature

The `POST` method could have the following signature:

```typescript
post: (postUrl: string, postData: any) => Promise<void>
```

- `postUrl`: The URL to which the data should be posted.
- `postData`: The data that should be posted to the specified URL.

## Example Use-Case

Here is a sample use-case demonstrating how to use the `POST` method.

```jsx
import { useFetch } from './useFetch';

const url = `http://jsonplaceholder.typicode.com/posts`;

interface Post {
  userId: number;
  id: number;
  title: string;
  body: string;
}

export default function Component() {
  const { data, post, error } = useFetch<Post[]>(url);

  const handleFormSubmission = async () => {
    const postData = { userId: 1, title: 'New Post', body: 'This is a new post.' };
    try {
      await post(url, postData);
      console.log('Form submitted successfully.');
    } catch (err) {
      console.error('Form submission failed:', err);
    }
  };

  if (error) return <p>There is an error.</p>;
  if (!data) return <p>Loading...</p>;

  return (
    <div>
      <p>{data[0].title}</p>
      <button onClick={handleFormSubmission}>Submit Form</button>
    </div>
  );
}
```

## Future Extensions

- Adding support for other HTTP methods like `PUT`, `DELETE`, etc.
- Adding the ability to customize headers for the `POST` request.

## Summary

Adding a `POST` method to the `useFetch` hook would make the hook more versatile and user-friendly. It would provide a unified and consistent API for handling both `GET` and `POST` HTTP methods, improving the developer experience.

## Todos

- [ ] Add unit testing
- [ ] Add a working example
- [ ] Update the documentation to reflect new features

### DIFF
--- a/packages/usehooks-ts/src/useFetch/useFetch.demo.tsx
+++ b/packages/usehooks-ts/src/useFetch/useFetch.demo.tsx
@@ -10,9 +10,31 @@ interface Post {
 }
 
 export default function Component() {
-  const { data, error } = useFetch<Post[]>(url)
+  const { data, post, error } = useFetch<Post[]>(url)
+
+  const handleFormSubmission = async () => {
+    // Mock-up data to be sent
+    const postData = {
+      userId: 1,
+      title: 'New Post',
+      body: 'This is the body of a new post',
+    }
+
+    try {
+      await post(url, postData)
+      console.log('Form submitted successfully.')
+    } catch (err) {
+      console.error('Form submission failed:', err)
+    }
+  }
 
   if (error) return <p>There is an error.</p>
   if (!data) return <p>Loading...</p>
-  return <p>{data[0].title}</p>
+
+  return (
+    <div>
+      <p>{data[0].title}</p>
+      <button onClick={handleFormSubmission}>Submit Form</button>
+    </div>
+  )
 }
--- a/packages/usehooks-ts/src/useFetch/useFetch.ts
+++ b/packages/usehooks-ts/src/useFetch/useFetch.ts
@@ -13,10 +13,23 @@ type Action<T> =
   | { type: 'fetched'; payload: T }
   | { type: 'error'; payload: Error }
 
+interface CustomFetchOptions {
+  customHeaders?: HeadersInit
+}
+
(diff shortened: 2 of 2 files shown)


juliencrn:
Hello @iamwillnbcu, 
Thanks a lot for this great work, what a PR description 👍
Unfortunately, we'll depreciate this hook.

PR #391, opened 2023-10-13
Fix module resolution

fedemp:
Split types for CJS and ESM.
Rename ESM files to .mjs so they are not mislabled as CJS. Add extension to imports to comply with node16 resolution mode.


mariusGundersen:
It would be great if this was part of the next release :heart:

juliencrn:
It's a bit more complicated and will require more investigation. 

But it will be in the next major release for sure!

PR #400, opened 2023-11-28
Feature: Add useEffectTill

samavati:
### Summary

This PR introduces a custom React hook, `useEffectTill`, designed to run an effect only until it is marked as done, handling cleanup afterward. The hook enhances the standard `useEffect` by allowing for precise control over when the effect should run and when it should be cleaned up.

### Changes Made

- Added TypeScript type definitions for the callback functions used in the hook.
- Implemented the `useEffectTill` hook, including tracking if the effect has already run and handling cleanup.
- Added internal comments to enhance code readability and understanding.

### Usage Example

```ts
useEffectTill((done) => {
  if (data) {
    // Do something.
    done();
  }
}, [data]);
```

### Why

The `useEffectTill` hook fills a gap in the React ecosystem by offering developers a way to execute effects with precise control over their lifecycle. It enhances code readability and reduces potential bugs related to effect cleanup.

### How to Test

1. Apply the `useEffectTill` hook in a React component.
2. Write tests to cover scenarios where the effect should and should not run.
3. Verify that cleanup occurs correctly once the effect is marked as done.

### Additional Notes

- This implementation aligns with best practices in React development.
- The code includes appropriate TypeScript typings for a seamless integration into TypeScript projects.

### DIFF
--- a/packages/usehooks-ts/src/useEffectTill/useEffectTill.demo.tsx
+++ b/packages/usehooks-ts/src/useEffectTill/useEffectTill.demo.tsx
@@ -1,0 +1,27 @@
+import { useEffect, useState } from 'react'
+
+import { useEffectTill } from './useEffectTill'
+
+export default function Component() {
+  const [data, setData] = useState<number>(0)
+  useEffect(() => {
+    console.log('Normal useEffect', { data })
+  }, [data])
+
+  useEffectTill(
+    done => {
+      if (data === 0) {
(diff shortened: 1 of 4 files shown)


casoer228 (approved):

sambauers:
I think `useEffectUntil` is a better name if this is being added. "Til" is a contraction of "until" anyway (and mis-spelt as "till" here).

PR #426, opened 2024-01-17
useReadLocalStorage: localStorage failing to parse when not JSON

nateforsyth:
I've updated useReadLocalStorage to attempt to return T after failing to parse JSON when localStorage values aren't stored as JSON.

Console output.
image

Debug, plain string localStorage item.
image

Error handler.
image


juliencrn:
Should be fixed by #439

Issue #453, opened 2024-02-01
[BUG] Can't resolve `lodash.debounce`


eleazareramos:
### Describe the bug

Latest version `2.11.0` introduced this build error:

```
Failed to compile.
--
19:16:17.545 |  
19:16:17.545 | ./node_modules/usehooks-ts/dist/esm/useDebounceCallback/useDebounceCallback.js
19:16:17.545 | Module not found: Can't resolve 'lodash.debounce'
19:16:17.545 |  
19:16:17.545 | https://nextjs.org/docs/messages/module-not-found
19:16:17.545 |  
19:16:17.545 | Import trace for requested module:
19:16:17.545 | ./node_modules/usehooks-ts/dist/esm/useDebounceCallback/index.js
19:16:17.545 | ./node_modules/usehooks-ts/dist/esm/index.js
```

### To Reproduce

Build error occurred on a NextJS app at version `13.5.2` during deployment to Vercel.
I also have `lodash` in my `package.json` at version `4.17.21`

### Expected behavior

`lodash.debounce` module to be resolved

juliencrn:
Hi, `lodash.debounce` is a usehooks' optional peer dependency required in `useDebounceCallback`. You need to `npm i lodash.debounce` to use this hook. It was added that way to minimize the bundle size, only pay for what you use.

However, it could be interesting to fall back to `lodash` if `lodash.debounce` is not available, idk if it's possible.

PR #459, opened 2024-02-01
fix: move `lodash.debounce` to `dependencies` from `peerDependencies`

### LINKED ISSUE #458: [BUG] Missing modules whie using with Next.js
### Describe the bug

I am trying to use `useInterval` hook with Next.js App router, and I am getting a error `Module not found: Can't resolve 'lodash.debounce'` while trying to do so.

The Error is specfic to `usehooks-ts 2.11.0`

image

### To Reproduce

1. Create new Next.js App with defaults
2. Install `usehooks-ts 2.11.0`
3. Try to use `useInterval` hook in a client component

### Expected behavior

It should work as it did in previous versions, also the error comes from `useDebounceCallback` which I am not even importing.

### Additional context

_No response_

BlankParticle:
Moved `lodash.debounce` to `dependencies` from `peerDependencies`.

Fixes #458, #452, #453

More info about the bug in #458

### DIFF
--- a/.changeset/pink-ravens-fix.md
+++ b/.changeset/pink-ravens-fix.md
@@ -1,0 +1,5 @@
+---
+"usehooks-ts": patch
+---
+
+move `lodash.debounce` to dependencies from peerDependencies
--- a/packages/usehooks-ts/package.json
+++ b/packages/usehooks-ts/package.json
@@ -48,14 +48,11 @@
     "vite": "^5.0.12",
     "vitest": "^1.2.2"
   },
-  "peerDependencies": {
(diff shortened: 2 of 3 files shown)


timlindsk (approved):

juliencrn (approved):

juliencrn:
Thanks @BlankParticle! 

It's a temporary fix btw, indeed `lodash.debounce` is very small, but, I'd be happy to make it work to be able to release more hooks with deps (`useDebounce` using `lodash.debounce`, `useCookie` using `cookie-js` etc)

If someone wants to work on it, it's more than welcome!

PR #463, opened 2024-02-02
chore: remove vite and use vitest only

BlankParticle:
This PR removes `vite` and `@vitejs/plugin-react` as dependencies.
Vite is not used in the project and only `vitest` is used for testing. So its better to have a `vitest.config.mts` for that purpose only. Having `vite` is confusing as it is not even used.

### DIFF
--- a/packages/usehooks-ts/package.json
+++ b/packages/usehooks-ts/package.json
@@ -37,7 +37,6 @@
     "@types/lodash.debounce": "^4.0.9",
     "@types/node": "^18.7.18",
     "@types/react": "18.2.48",
-    "@vitejs/plugin-react": "^4.2.1",
     "date-fns": "^3.3.1",
     "eslint-config-custom": "workspace:*",
     "eslint-plugin-jsdoc": "^48.0.4",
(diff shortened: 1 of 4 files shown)


BlankParticle:
Also, on a side note, I think you should consider to move to `tsup` and single file exported library like most other libs out there.

Currently you have a lot of folders in `dist` which re-export everything to the outermost `index.js`, instead you can just pack all of the hooks into one giant `index.mjs`, `index.js` and `index.d.ts`.

`tsup` takes care of all this by default, also the build time drops from `10s` to `3s` on my machine.

As always, I can take care of this if you want.

juliencrn (approved):
Nice! Thank you 👍

juliencrn:
> Also, on a side note, I think you should consider to move to tsup and single file exported library like most other libs out there.

> Currently you have a lot of folders in dist which re-export everything to the outermost index.js, instead you can just pack all of the hooks into one giant index.mjs, index.js and index.d.ts.

> tsup takes care of all this by default, also the build time drops from 10s to 3s on my machine.

> As always, I can take care of this if you want.

Yes, you're right. I already did some tests about that, it will be in the v3. I just want to process the opened issues/PRs first to batch all the breaking changes together in the major release

Issue #271, opened 2022-12-27
useIntersectionObserver received 2 entries, but only the first (outdated) is taken


reinos:
Hi,

First of all, what a great library of hooks, well done!

Currently I have an issue with the intersectionObserver. In Chrome (not in FF nor Safari) I receive 2 entries in the callback. The first entry has `isIntersecting : false` and the second has `isIntersecting: true`. Im not sure why I receive 2 entries via the Intersection callback, but it seems that the first one is obsolete (outdated) and the latest one is the most up to date with the DOM.

So I feel that the implementation here https://github.com/juliencrn/usehooks-ts/blob/master/src/useIntersectionObserver/useIntersectionObserver.ts#L20-L22 of just getting the first entry is wrong. I think this should just catch the latest one. As it seems there are situations where the `Intersection callback` can give more than one result and where latests one is always the one that is up-to-date.

qinbinhua601:
@reinos I somehow encountered the same problem occasionally but could not find the 100% reproduce step. Would you please share the reproduce step on how to reproduce it in a minimal demo. Thanks in advance. 😄 😄 😄

juliencrn:
Hi, thanks for the feedback!
I've pushed a PR that should fix it,  feel free to give a code review on it.
See #464

Issue #182, opened 2022-07-14
React 18: useIntersectionObserver loses binding when ref conditionally rendered


jdpt0:
In the below example I'm passing in an infinite react-query hook into this component to load more results once the bottom of the list has been reached. With `useIntersectionObserver` this happens the first time, but after that, the binding is lost and the `Load more` button has to be clicked to initiate the fetching of the query. Tested on React 17 and works as expected.

A better pattern might be to pass a ref from the hook to bind to a specific component, as is done [here](https://github.com/thebuilder/react-intersection-observer/blob/master/src/useInView.tsx).

```javascript
export const InfiniteList = ({
  query,
  children,
  className,
  style,
}: {
  query: UseInfiniteQueryResult;
  children: ReactNode;
  className?: string;
  style?: CSSProperties;
}): ReactElement => {
  const ref = useRef<HTMLButtonElement | null>(null);

  const entry = useIntersectionObserver(ref, {});
  const isIntersecting = !!entry?.isIntersecting;

  useEffect(() => {
    if (
      isIntersecting &&
      query.data &&
      !query.isLoading &&
      !query.isFetchingNextPage
    )
      query.fetchNextPage();
  }, [query, isIntersecting]);

  const renderLoadButton = () => {
    if (!query || !query.hasNextPage) return null;
    if (query.isFetchingNextPage) return <Spinner />;
    return (
      <div className={styles.loadMoreContainer}>
        <Button size={"small"} ref={ref} onClick={() => query.fetchNextPage()}>
          Load more
        </Button>
      </div>
    );
  };

  return (
    <div className={classNames("overflow-auto", className)} style={style}>
      {children}
      <div className={styles.loadMoreContainer}>{renderLoadButton()}</div>
    </div>
  );
};

parkerault:
This is just because the dependency array in the useEffect in `useIntersectionObserver` is using the ref instead of ref.current. I just ran into this and changed the dependency array and it works as expected now. I will submit a PR in a bit.

jbean96:
This still seems to be a problem for me even after the fix that changes the dep array for `useIntersectionObserver` to `elementRef?.current`.

electroheadfx:
I have too this issue, seem a problem related with react 18 (and Next ? I use SSR), not sure. The ref is lost. It works on a page and not in anothers, I need to investigate more to understand the issue ...

scottrippey:
[...]

PR #464, opened 2024-02-02
✨ Improve useIntersectionObserver

### LINKED ISSUE #395: useIntersectionObserver return previous entry as well
It seems like it'd be generally useful to have the previous entry (or list of prev entries) returned from the hook. That way client code could compare timings, e.g.

```typescript
const [ entry, prevEntry ] = useIntersectionObserver(ref, {})
const timeOnScreen = prevEntry ?? entry.time - prevEntry.time
```

or all previous entries like:

```typescript
const { entry, previousEntries } = useIntersectionObserver(ref, {})
// ... do fancy comparisons
```

juliencrn:
- Updated the hook API (its signature) using function overloading to keep both the old and the new interface working together. The old one has been depreciated and will be removed in the incoming v3
  - from `const entry = useIntersectionObserver(ref, options)`
  -  to `const [ref, isIntersecting, entry] = useIntersectionObserver(options)` 
- Support both object and tuple destructuring for the hook return 
  - `[ref, isIntersecting, entry]`
  - `{ ref, isIntersecting, entry }`
- Refresh when the `ref` changes to always have the latest
- Added `onChange` optional callback
- Added `initialIsIntersecting` optional flag

This PR should fix #395, #271 and #182. 
Open to code review!


BlankParticle (approved):
I tested the functionality of the hook, and things seems to work as expected

packages/usehooks-ts/src/useIntersectionObserver/useIntersectionObserver.ts
```diff
@@ -1,23 +1,65 @@
...
+  ref: (node?: Element | null) => void
+  isIntersecting: boolean
+  entry?: IntersectionObserverEntry
 }
 
 /**
+ * Custom hook for tracking the intersection of a DOM element with its containing element or the viewport.
+ * @param {IntersectionObserverOptions} options - The options for the Intersection Observer.
+ * @returns {IntersectionResult} The ref callback, a boolean indicating if the element is intersecting, and the intersection observer entry.
+ *  * @see [Documentation](https://usehooks-ts.com/react-hook/use-intersection-observer)
```
juliencrn, line 46:
```suggestion
 * @see [Documentation](https://usehooks-ts.com/react-hook/use-intersection-observer)
```

PR #468, opened 2024-02-04
✨ Feature: add `useSound` hook

BlankParticle:
Adds a new hook called `useSound` that allows you to play a sound effect.

## Todo

- [x] `useSound` hook
- [x] Docs
- [x] Example
- [ ] Tests

PS: I don't have any experience with testing, I am not sure how to test this hook. If you have any idea, please let me know. I will add tests for this hook.


BlankParticle:
I think we are good to merge this, We can take care of tests later. The hook works as expected already.

juliencrn:
Hey @BlankParticle,
Good idea, but it's incomplete for now, there are some things to update: 
- Instantiating the `Audio` element on mount fails in SSR context, you can reproduce that error trying the example in NextJS, you could use a ref and either initialize it straight in CSR and in an `useEffect` in SSR (ctrl + F `initializeWithValue` for examples)
- I think we should clean it (stop the player, remove it) on unmount, see: https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement/Audio#memory_usage_and_management
- We maybe should provide an option or mechanism to wait for the player to be ready
- It could be very nice to expose a broader API, maybe returning the `audioRef`
- Passing explicit `undefined` in options like `Object.assign(new Audio("/bob.mp4"), { volume: undefined })` fails

BlankParticle:
These are certainly valuable feedback, I will implement them asap.

BlankParticle:
I fixed the issue with SSR, only `preload` can be passed as extra arg, other options can be chnaged on the exposed `ref`, `audioRef` is paused and `null`ed out on effect clean up for garbage collection.

About a mechanism for waiting for player to be ready, I tried this, do you think its a good idea?

```ts
const waitTillPlayerIsReady = (timeout?: number) => {
    if (!audioRef.current) return Promise.resolve()
    return new Promise<void>((resolve, reject) => {
      const onCanPlay = () => {
        if (!audioRef.current) return
        audioRef.current.removeEventListener('canplay', onCanPlay)
        resolve()
      }
      audioRef.current?.addEventListener('canplay', onCanPlay)
      audioRef.current?.addEventListener('error', reject)

      if (timeout) {
        setTimeout(reject, timeout)
      }
    })
  }
  ```
I can expose it as a function on the result object

PR #469, opened 2024-02-04
🎉 feat: Improve the site

BlankParticle:
Various improvements to the site
Each commit expalins their purpose, I didn't want to make separate PRs for everything.

I already have enough PRs open 😅


juliencrn:
Nice @BlankParticle, thank you! 

I just removed the images from your PRs because it does not follow the design.

Additionally, I also removed the switch color mode to only use the user preference, it allows us to simplify the code and to dynamically load themed code highlight

PR #471, opened 2024-02-05
chore: Improve docs introduction

BlankParticle:
I have started the work with the new `CommandCopy` component. 
~I have wrote 2 lines for now. You may write up the rest. I will check on later.~

I have written the introduction, You may change up the wording if you want.


juliencrn:
Cool 👏 Thank you. 

- Not sure about the package manager dropdown UX, we don't know about the select options before clicking, but nice though.
- This page (and the possible future other doc pages too) looks like a good use-case for MDX (`www/src/app/(docs)/introduction/page.mdx`) to reduce the boilerplate. 

I agree that this page was pretty empty 😅 It's much better right now, thanks 👍

BlankParticle:
> Cool 👏 Thank you.
> 
> * Not sure about the package manager dropdown UX, we don't know about the select options before clicking, but nice though.
> * This page (and the possible future other doc pages too) looks like a good use-case for MDX (`www/src/app/(docs)/introduction/page.mdx`) to reduce the boilerplate.
> 
> I agree that this page was pretty empty 😅 It's much better right now, thanks 👍

I also don't like that UX that much, That drop down is almost one to one copy of shadcn's command copy. I think We can change that in a major site redesign later

PR #479, opened 2024-02-09
✨ Feature: add `useScrollLock` hook

BlankParticle:
Adds a new hook called `useScrollLock` that allows you to automatically/programmatically lock the scroll of target element.

## Todo

- [x] `useScrollLock` hook
- [x] Docs
- [x] Example
- [x] Tests


juliencrn:
Duplicate of `useLockedBody`? However, it brings some interesting stuffs

BlankParticle:
The most usefull feature is auto lock and auto unlock, its based on <https://usehooks.com/uselockbodyscroll> but has a larger feature set.

I think this can also replace `useLockedBody` if you want be to implement scrollbar width reflow hack

BlankParticle:
@juliencrn, Hey I was busy with exams and didn't check much, I see that the hooks I made PRs for are still sitting unreviewed. Can you please review and accept the PRs in mean while. I would be free from tommorow, when I could work more on the project.

Thanks

juliencrn:
Hi @BlankParticle,
I made some adjustments to the hook:
- remove the internal state
- add JSDoc param details
- add width reflow support
- simplify a bit the useEffect
- make the demo working
- remove eslint-disable comment

What do you think about it?

BlankParticle:
Width reflow should be an option that is enabled by default but can be disabled if the user wants

juliencrn:
We also should remove this file `apps/www/src/hooks/use-lock-body.ts` used in `apps/www/src/components/mobile-nav.tsx` and use the new one from `usehooks-ts` instead

BlankParticle:
> We also should remove this file `apps/www/src/hooks/use-lock-body.ts` used in `apps/www/src/components/mobile-nav.tsx` and use the new one from `usehooks-ts` instead

I will do that in a new PR

BlankParticle:
Oh, you did the work on site already. I think we are good to merge 🚀

juliencrn:
Yes totally, it's more than welcome 🙌 Thank you

PR #484, opened 2024-02-10
feat: add onStop to useCountdown hook

samyarkd:
Every time I used useCountdown I had to set a timeout for when the count is stopped, it would much easier if we have this inside the hook, it would be much cleaner.

### DIFF
--- a/packages/usehooks-ts/src/useCountdown/useCountdown.ts
+++ b/packages/usehooks-ts/src/useCountdown/useCountdown.ts
@@ -23,6 +23,7 @@ interface CountdownOptions {
   intervalMs?: number
   isIncrement?: boolean
   countStop?: number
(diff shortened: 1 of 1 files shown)


samyarkd:
please let me know if I have to add jsDoc to the hook in case of approval.

juliencrn:
Hi @samyarkd,
Thanks for that PR. 
Yes, the JSDoc is more than welcome, and a unit-test too. 
However, the PR does not compile yet.

samyarkd:
added a unit test and modified the jsDocs 👍 also created a function to stop the countdown.

samyarkd:
@juliencrn congratulations on the V3, i noticed there has been a couple changes in the hook so modified the changes according to the V3 so there are no conflicts

PR #487, opened 2024-02-16
feature: add a reset function to useLocalStorage

lindesvard:
Added a `reset` function to `useLocalStorage`. I want to reset the state to the initial value. I know this is feasible without this change but thought it would be a nice addition.

Up to you guys if you want to merge it.


BlankParticle (changes requested):
The concept is ok, but we shouldn't expose a third variable in the function. You should keep the output array as it is, as it resembles useState.
You can make the reset function available on object destructuring.

Check `useIntersectionObserver` for example. If you don't understand something, feel free to ask.

PR #488, opened 2024-02-17
✨ [Feature] :added new  hooks  `useHash`, ` useHashHistory`, `useClipboard ` and  `useOrigin`

haseeb5555:
-key changes 

four new hooks  ✔️
docs ✔️
test cases✔️


BlankParticle (changes requested):
Overall, 5 out of 7 test cases are failing. Every hook is written differently. Test cases and demos don't make sense for some hooks. Also there are 4 hooks in this 1 PR that shouldn't be the case.

packages/usehooks-ts/src/useOrigin/useOrigin.ts
```diff
@@ -0,0 +1,18 @@
...
+    setMounted(true);
+  }, []);
+
+  const origin = typeof window !== "undefined" && window.location.origin ? window.location.origin : "";
+
+  if (!mounted) {
+    return "";
+  }
+
+  return origin;
```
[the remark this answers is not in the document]
haseeb5555: Got it thanks for feedback. Will fix it soon

packages/usehooks-ts/src/useHash/useHash.ts
```diff
@@ -0,0 +1,21 @@
...
+        setHash(window.location.hash.slice(1));
+      };
+      handleHashChange();
+
+      window.addEventListener("hashchange", handleHashChange);
+
+      return () => {
+        window.removeEventListener("hashchange", handleHashChange);
+      };
+    }, []);
```
jiji-hoon96, line 18: If I modify the code so that it renders anew every time the value of the hash changes, will this cause problems?

jiji-hoon96 (commented):
useOrigin and useHashHistory are so interesting.

packages/usehooks-ts/src/useHash/useHash.ts
haseeb5555: Modifying the code to re-render on hash changes shouldn't inherently cause problems. It's done to ensure the component reacts to hash value changes accurately.

PR #494, opened 2024-02-20
Fix setMatches hook in useMediaQuery by defining getMatches before use

ymc-maha:
This PR fixes the setMatches hook in useMediaQuery in version 2.14.0. 

If initializeWithValue is set to true getMatches is called before it is defined. This issues comes with v. usehooks-ts@2.14.0
https://github.com/juliencrn/usehooks-ts/commit/4a9fc884c3e913f7d4c78739d2e4184ece2c4d72

### DIFF
--- a/.changeset/real-foxes-fold.md
+++ b/.changeset/real-foxes-fold.md
@@ -1,0 +1,5 @@
+---
+"usehooks-ts": major
+---
+
+Fix setMatches hook in useMediaQuery by defining getMatches before use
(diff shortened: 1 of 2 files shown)


BlankParticle (approved):
Kudos for catching that, I think we might need to add tests for all hooks that deals with SSR 👀

juliencrn (approved):
Good catch, thanks!

juliencrn:
@all-contributors please add @ymc-maha for code and bug

PR #498, opened 2024-02-20
chore: delete extra `.d.ts` file to fix duplicate declarations

BlankParticle:
As we support both `esm` and `cjs` versions, `tsup` generates `.js` and `.mjs` files along with `.d.ts` and `.d.mts` files. Even though only one `.d.ts` file is needed. This leads to duplicated entries in the auto complete.
image

I haven't found any way to disable this behaviour and have one `.d.ts` file for multiple output formats. So I have added an extra `rm` in build command to fix the issue for now. I will make an issue in `tsup`'s repo about this. After a fix is found, I will add another PR addressing this.

### DIFF
--- a/packages/usehooks-ts/package.json
+++ b/packages/usehooks-ts/package.json
@@ -23,7 +23,7 @@
   },
   "sideEffects": false,
   "scripts": {
-    "build": "tsup",
+    "build": "tsup && rm ./dist/index.d.mts",
     "dev": "tsup --watch",
     "test": "vitest run",
     "test:watch": "vitest",


juliencrn:
I'm not able to reproduce it. Tested using `usehooks-ts@latest` in Nextjs typing `useB`. What about your environment?

BlankParticle:
I tried with both vite and nextjs, also tried VSCode and NeoVim

image

I happens when you try to use the fucntion without importing it

juliencrn:
Thanks!

Issue #423, opened 2024-01-16
Hook unit-testing journey


juliencrn:
## Overview

This GitHub issue serves as a central hub for the unit-testing journey of our React hook library. Our goal is to ensure robust and reliable testing for each individual hook in the library.

## Objectives

1. **Comprehensive Testing**: Write unit tests for each hook to ensure thorough coverage of functionality.
2. **Consistent Test Structure**: Maintain a consistent structure/format for unit tests across all hooks.
3. **Documentation**: Document the purpose and usage of each test to enhance overall project understanding.

## Getting Started

1. Fork the repository to your account.
2. Create a new branch for your tests: git checkout -b feature/hook-name-tests.
3. Write tests for the specific hook in `packages/usehooks-ts/src/useExample/useExample.test.ts`.
4. Ensure all tests pass before submitting a pull request.

## Hooks to Test

- [x] `useBoolean`
- [ ] `useClickAnyWhere`
- [x] `useCopyToClipboard`
- [x] `useCountdown`
- [x] `useCounter`
- [x] `useDarkMode`
- [x] `useDebounce`
- [x] `useDocumentTitle`
- [x] `useEffectOnce`
- [x] `useElementSize`
- [ ] `useEventCallback`
- [x] `useEventListener`
- [ ] `useFetch`
- [x] `useHover`
- [ ] `useImageOnLoad`
- [ ] `useIntersectionObserver`
- [x] `useInterval`
- [x] `useIsClient`
- [x] `useIsFirstRender`
- [x] `useIsMounted`
- [ ] `useIsomorphicLayoutEffect`
- [x] `useLocalStorage`
- [ ] `useLockedBody`
- [x] `useMap`
- [ ] `useMediaQuery`
- [ ] `useOnClickOutside`
- [x] `useReadLocalStorage`
- [ ] `useScreen`
- [ ] `useScript`
- [x] `useSessionStorage`
- [ ] `useSsr`
- [x] `useStep`
- [ ] `useTernaryDarkMode`
- [x] `useTimeout`
- [x] `useToggle`
- [x] `useUpdateEffect`
- [x] `useWindowSize`

Let's ensure our hooks are well-tested and reliable!

PR #499, opened 2024-02-20
chore: fix `isTestFile` in utils to include `.tsx` test files

BlankParticle:
Updates `isTestFile` function in utils.mjs to include `.tsx` test files also.

PS: Update #423 after this change


juliencrn (approved):
Good catch, thanks

PR #501, opened 2024-02-21
improvement: omit useDebounceCallback func arg from dependency array

therour:
## Improvement
as user's provided `func` maybe not necessarily memoized, so to avoid re-evaluation of `debounced` we can just store the `func` inside ref, and remove it from `useMemo` dependency array

same goes with the `options`, the user's provided `options` object also not necessarily memoized, so we can just pass it's values to dependency array instead of the object reference itself

### Current behavior
`isPending()` always return `true`

### Expectation
based on type comments
```
/**
   * Checks if there are any pending function invocations.
   * @returns `true` if there are pending invocations, otherwise `false`.
   */
  isPending: () => boolean
```


juliencrn:
Hi @therour, your PR does not compile yet.

therour:
@juliencrn thanks for the response,
actually I got the ts error `Type 'Parameters<T>' must have a '[Symbol.iterator]()'` 

I think the Generic for DebouncedState should be changed to `<T extends (...args: any[]) => ReturnType<T>>` based on  https://github.com/microsoft/TypeScript/issues/41728

but when I changed this in this PR, this will break another part that uses this interface

therour:
OK I understand, that it can be fixed from developer perspective by memoize the callback arg

```js
const debouncedIncrement = useDebounceCallback(
    React.useCallback(() => setValue(val) => val + 1, []), 
    500
)
```

PR #503, opened 2024-02-22
🔥 Remove deprecated hooks and co

juliencrn:
Use the major release to batch all breaking changes in the hooks. 

This PR removes previously deprecated hooks and old hooks' signatures.

### Removed

- `useDebounce`: Replaced by both `useDebouceValue` and `useDebounceCallback`.
- `useEffectOnce`: unnecessary abstraction.
- `useElementSize`: Replaced by more performant `useResizeObserver`.
- `useFetch`: Prefer RSC, framework methods, React's `use`, React-query, useSWR...
- `useImageOnLoad`: Too opinionated.
- `useIsFirstRender`: Not comply with the React functional mindset (cf: React 18 render twice in strict mode).
- `useLockedBody`: Replaced by the more generic `useScrollLock`.
- `useSsr`: It was not a React hook, other hooks can do the job.
- `useUpdateEffect`: unnecessary abstraction.

### Hooks with dropped legacy signature

- `useCountDown`
- `useDarkMode`
- `useIntersectionObserver`
- `useMediaQuery`
- `useTernaryDarkMode`

### Additional

Also updated the website to allow hook pages without `code` or `demo`. Later, we could consider removing the doc for removed hooks and use permanent redirection to the alternative.


BlankParticle:
Before v3 remember to merge the automated workflow PR.

PR #504, opened 2024-02-22
fix useResizeObserver initialSize mutation

iuriiiurevich:
This PR fixes a mutation of the initialSize, due to which the hook may not work correctly when there is two or more instances of it.


juliencrn (dismissed):
Good catch, thank you 👍

juliencrn:
@all-contributors please add @iuriiiurevich for bug and code

PR #515, opened 2024-02-27
♻️ Refactor/prefer type over interface

juliencrn:


adidahiya:
@juliencrn what was the motivation for this change? Just curious.

juliencrn:
> @juliencrn what was the motivation for this change? Just curious.

Hi, initially, I was looking for uniformization as we used both interchangeably. We opted for `type` to reduce possible weird behavior, which is explained here: https://www.youtube.com/watch?v=zM9UPcIyyhQ. It's not very important though

adidahiya:
@juliencrn thanks for the video link, that was an interesting watch.

It looks like you're a solo maintainer of this project, so PR documentation might not be a top priority for you (I've been in the same position), but as a user of the library and watcher on Github, I would love it if your PRs had non-empty descriptions so that I could keep track of changes a little better :) thanks

PR #516, opened 2024-02-29
fix: Changed order in lock() function to adjust padding before changing the overflow

novacdenis:
Changed the order because when we set the overflow to hidden, the scrollbar disappears; therefore, the scrollbar width is always zero. I couldn't reproduce it in a test because the scrollbar widths are always 0 inside a test.

https://github.com/testing-library/react-testing-library/issues/353


juliencrn:
Hi @novacdenis,
As I merged #521 before, this PR now only includes test changes.
Thank you!

Issue #520, opened 2024-03-03
[BUG] useScrollLock - Do not leave inline styles if they weren't present before the lock


kyrylo-soulandwolf:
### Describe the bug

1. At the moment the hook leaves `overflow` and `padding-right` on the target after unlocking. Even though it is a computed value from before the lock, it can change in CSS for various reasons (and not considered in future locks) and will always be overridden by inline styles. These 2 values should be reverted to the initial value if they existed inline, otherwise removed completely which will make the styles on the target the same as before the lock.

[Here's an example](https://github.com/tuax/tua-body-scroll-lock/blob/master/src/index.ts#L35) how `tua-body-scroll-lock` does this. 

2. The second problem is that [the global scrollbar is not part of the body](https://stackoverflow.com/a/68547183/19714283), so if `lockTarget` is a `body` element `scrollbarWidth` will always be 0. There should be a special case for `body` as a target and use `window.innerWidth` instead of `target.current.offsetWidth` to get the width with the global scrollbar.

### To Reproduce

[Codesandbox repro link](https://codesandbox.io/p/sandbox/usehooks-ts-usescrolllock-test-dll46t?file=%2Fsrc%2FApp.tsx%3A10%2C12)

1. Just lock and unlock any element.
2. Same thing but using `body` as a target

### Expected behavior

After unlocking, the target element should have initial inline styles or no inline styles if they weren't present.

kyrylo-soulandwolf:
The tests seem to pass by returning false positives because `window.getComputedStyle(...)` returns an empty string instead of the actual value

kyrylo-soulandwolf:
Opened a pull request https://github.com/juliencrn/usehooks-ts/pull/521 😊

PR #521, opened 2024-03-03
Fix useScrollLock leaving inline styles and not considering initial padding before the lock

kyrylo-soulandwolf:
Fixes https://github.com/juliencrn/usehooks-ts/issues/520

### DIFF
--- a/.changeset/fresh-emus-nail.md
+++ b/.changeset/fresh-emus-nail.md
@@ -1,0 +1,5 @@
+---
+"usehooks-ts": patch
+---
+
+Add isLocked state to the hook return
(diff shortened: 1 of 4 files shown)


kyrylo-soulandwolf:
I've also included a fix from https://github.com/juliencrn/usehooks-ts/pull/516 which fixes the order of locking the element and calculating the padding
@novacdenis tagging you so you're also aware :wink:

kyrylo-soulandwolf:
Also, added `isLocked` state and appended it to the hook return which could be useful when locking and unlocking manually.

juliencrn (approved):
LGTM, thanks 👍

PR #550, opened 2024-03-21
feat(useEventCallback): allow optional callback

Newbie012:
Updated the useEventCallback hook to accept an optional callback function. This improves developer experience by allowing more flexibility in the use of the hook.

This is quite useful when you want to memoize a callback prop that can be optional. For instance:

```ts
type Props<T> = {
  onChange?: (value T | undefined) => void;
}

function Comp({ onChange }: Props) {
  const onChangeRef = useEventCallback(onChange); 

  return <ExpensiveCompImpl onChange={onChangeRef} />
}
```


juliencrn:
Hi @Newbie012, well done, thank you 👍

juliencrn:
@all-contributors please add @Newbie012 for code

Issue #512, opened 2024-02-23
useDarkMode resets


vhpoet:
### Describe the bug

[This](https://github.com/juliencrn/usehooks-ts/blob/master/packages/usehooks-ts/src/useDarkMode/useDarkMode.ts#L79-L84) is causing the `isDarkMode` to reset to the OS setting on every refresh. For some reason it works on production though.

### To Reproduce

Use it in local dev with next.js where the OS is set to dark mode.

### Expected behavior

I expect my choice to be persisted.

smith558:
`useDarkMode` currently not working on Linux either (Ubuntu), resets on every page load.

juliencrn:
Hi all, thanks for the bug report, feel free to open a PR, I will check it in priority

refeals:
Just created https://github.com/juliencrn/usehooks-ts/pull/557

The change in `useLocalStorage` may possibly break other hooks, but I tried to make it consistent. It still does not return `undefined` when the key does not exist, but now it creates the key/value in local storage when the hook is called. That asserts the local storage as the source of truth.

And that should fix the `useDarkMode` inconsistency too, so I removed the `useIsomorphicLayoutEffect` since it's now unnecessary - the `isDarkMode` variable is always reflecting the correct data in local storage and it does not update on refresh.

crucialfelix:
This PR fixes it for me — was always reverting to the OS default on every page load.

gregmsanderson:
Ah, been finding this too. The chosen option is lost on a refresh as localStorage is reset. That explains why!

pavitra-infocusp:
I have tripped on this issue too.

iavorskyi:
Had the same problem. Fixed with this approach too. Looking forward to this fix will be delivered to the main.

> I have tripped on this issue too.
> 
> Fixed by copying the hook and commenting out `useIsomorphicLayoutEffect`
> 
> ```ts
> import {
>     useIsomorphicLayoutEffect,
>     useLocalStorage,
>     useMediaQuery,
>   } from 'usehooks-ts'
>   
>   const COLOR_SCHEME_QUERY = '(prefers-color-scheme: dark)'
>   const LOCAL_STORAGE_KEY = 'usehooks-ts-dark-mode'
>   
>   type DarkModeOptions = {
>     defaultValue?: boolean
>     localStorageKey?: string
>     initializeWithValue?: boolean
>   }
>   
>   type DarkModeReturn = {
>     isDarkMode: boolean
[...]

PR #557, opened 2024-03-25
fix: update useDarkMode to not reset with OS value on every refresh

refeals:
Fixes https://github.com/juliencrn/usehooks-ts/issues/512

This PR does two things:

### 1. Update useLocalStorage save initial value

The way things are now, useLocalStorage does not correctly represents the value in the `localStorage`. It returns a default value even if the `key` doesn't exist there yet and does not create it, meaning it returns a value that is not necessarily the truth (should be `undefined`. This causes a problem in `useDarkMode` (and probably other hooks).

The PR updates the hook to actually save the key/value in `localStorage` when the hook is called, preserving the hook's initial funcionality.

### 2. Update `useDarkMode` to not reset with OS value on every refresh

Removes the `useIsomorphicLayoutEffect` because it's unnecessary. The value `isDarkMode` will always be correctly set.

### DIFF
--- a/packages/usehooks-ts/src/useLocalStorage/useLocalStorage.ts
+++ b/packages/usehooks-ts/src/useLocalStorage/useLocalStorage.ts
@@ -145,7 +145,9 @@ export function useLocalStorage<T>(
   })
 
   useEffect(() => {
-    setStoredValue(readValue())
+    const value = readValue()
+    setValue(value)
+    setStoredValue(value)
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [key])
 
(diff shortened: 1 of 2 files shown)


crucialfelix:
This PR fixes it for me — was always reverting to the OS default on every page load.

refeals:
@juliencrn can you take a look?

blacklee123:
please merge this pr

rickyzhangca:
neeeeed

Rigidity:
Keep running into this issue repeatedly

abhijith-rubrik:
Can someone review and merge this

refeals:
Unfortunately @juliencrn is the only one how can approve and merge this :/

packages/usehooks-ts/src/useLocalStorage/useLocalStorage.ts
```diff
@@ -145,7 +145,9 @@ export function useLocalStorage<T>(
   })
 
   useEffect(() => {
-    setStoredValue(readValue())
+    const value = readValue()
+    setValue(value)
+    setStoredValue(value)
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
jgaik, line 151: setValue already sets stored value, isnt it?
maybe we can remove the eslint now and fill the deps as the logic changed?

Issue #549, opened 2024-03-19
[BUG] useLocalStorage's setValue should be callable in render phase (like a normal `useState`)


jraoult:
### Describe the bug

Calling the setter of  `useLocalStorage` in a render phase triggers the error with the message `Cannot call an event handler while rendering.`. It is because `useEventCallback` is used internally, but I think `useCallback` should be used instead since it is not conceptually only an event handler.

### To Reproduce

See the [reproduction sandbox](https://stackblitz.com/edit/vitejs-vite-keuhc5?file=src%2FApp.tsx), but the gist of it is:

```tsx
import { useEffect, useState } from 'react';
import { useLocalStorage } from 'usehooks-ts';

function App() {
  const [count, setCount] = useState(0);
  const [lsCount, setLsCount] = useLocalStorage('lsCount', 0);

  if (count === 0) {
    setCount(1);
  }

  // if (lsCount === 0) {
  //   // this should work but fails right now
  //   setLsCount(1);
  // }

  // this is the workaround
  useEffect(() => {
    if (lsCount === 0) {
      setLsCount(1);
    }
  }, [lsCount]);
```

### Expected behavior

It should be possible to set a value in the render phase.

juliencrn:
While I agree with the fact that it should be possible to set a value in the render phase, it's not a recommendable approach as the setter triggers a re-render every time the component is rendered. This can lead to performance issues and unexpected behavior, especially if you're updating state frequently or performing expensive operations.

This instead would be the right approach IMHO: 

```ts
  useEffect(() => {
    if (lsCount === 0) {
      setLsCount(1);
    }
  }, [lsCount]);
  ```

jraoult:
@juliencrn one last attempt to convince you:

>  the setter triggers a re-render every time the component is rendered.

This can also happen with `useEffect`. The way you schedule the update doesn't prevent the recursion. What matters is the bottom case ie. the `if (lsCount === 0)`.

>  it's not a recommendable approach

I think it is actually the recommended approach in the official React documentation (cf.  https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes). They recommend not using an effect and just setting the value in the render phase: "Start by deleting the Effect. Instead, adjust the state directly during rendering"

jraoult:
@juliencrn I also just wanted to add that if you change your mind, I'm happy to create a PR with the proper test coverage for this.

juliencrn:
[...]

PR #559, opened 2024-03-27
fix: fixed up bug(#549) -Init use `useState` don't work in `useLocalStorage` 

LonelyFellas:
#549

### DIFF
--- a/.changeset/late-teachers-hang.md
+++ b/.changeset/late-teachers-hang.md
@@ -1,0 +1,5 @@
+---
+"usehooks-ts": patch
+---
+
+fixed up bugs (#549)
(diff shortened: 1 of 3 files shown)


jraoult:
Thanks, @LonelyFellas, for trying to tackle that. That said, my gut feeling would be that the fix has to do with changing the `useLocalStorage.ts` file:

 https://github.com/juliencrn/usehooks-ts/blob/0f286bb13c28cb992c1c397ed1db22d4e72846b8/packages/usehooks-ts/src/useLocalStorage/useLocalStorage.ts#L122

to 
```ts
const setValue: Dispatch<SetStateAction<T>> = useCallback(value => {
```
With the appropriate dependency list rather than changing how `useEventCallback` behaves.

jraoult:
@LonelyFellas, just a disclaimer: I'm not a repo maintainer, so my opinion is super relevant here.

My first point is that `useCallback` is the proper hook in this context instead of`useEventCallback`. My point is that `useEventCallback` from (or precedes) the now abandoned [`useEvent` RFC](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md). It is explicitly built to stabilise the reference to a callback that is never run during a render phase. If a callback can/needs to be run in the render phase and we still want to stabilise the reference, we need to use `useCallback` instead (cf. [When useEvent should not be used?]( https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md#when-useevent-should-not-be-used)).

My second point would be that your proposed patch introduces a change in the contract of `useEventCallback`. Now `useEventCallback` return callback can be used in the render phase which it shouldn't  by design potentially breaking the code using it since the lack of dependency tracking. Here's a quote from the RFC:

> This makes it safe to preserve their identity despite the changing props/state inside. Because they can't be called during rendering, they can't affect the rendering output — and so they don't need to change when their inputs change (i.e. they're not "reactive").

But again, it would be up to @juliencrn to decide if my initial issue is something they think makes sense and then decide the proper way to implement it.

juliencrn:
Thanks for the proposition, but we'll abandon this change. (https://github.com/juliencrn/usehooks-ts/issues/549#issuecomment-2024680380).

juliencrn:
@all-contributors please add @jraoult for bug and review

juliencrn:
@all-contributors please add @LonelyFellas for code

PR #560, opened 2024-03-27
docs: improve `useCopyToClipboard` demo

luckrnx09:


packages/usehooks-ts/src/useCopyToClipboard/useCopyToClipboard.demo.tsx
```diff
@@ -3,14 +3,13 @@ import { useCopyToClipboard } from './useCopyToClipboard'
 export default function Component() {
   const [copiedText, copy] = useCopyToClipboard()
 
-  const handleCopy = (text: string) => () => {
-    copy(text)
-      .then(() => {
-        console.log('Copied!', { text })
-      })
-      .catch(error => {
```
luckrnx09, line 11: The `copy` method never throws a rejected Promise. 
We should let the user determine if the copy was successful by the resolved value.

juliencrn:
You're right, currently, the `copy` function never throws a error. But maybe instead of updating the doc, we should fix the hook to make the `copy` function "throwable" instead of returning a boolean. This way, we do have not to update the eslint rule and the hook's API will be more natural for JS dev. 
What do you think about?

luckrnx09:
Agree with you!
In fact, I prefer to throw an error if copy to clipboard was failed. 
Don't forget to tell the user if we do the breaking change. In particular, users who don't use TypeScript may not be aware that the return value of the copy function has changed from `Promise<boolean>` to `Promise<void>`.

luckrnx09:
Please feel free to close this PR

juliencrn:
Thank you for the quick response.  

I will keep this PR open until we update the hook as a reminder.

PR #561, opened 2024-03-27
fix: reject non-boolean value as a defaultValue for `useBoolean`

luckrnx09:


packages/usehooks-ts/src/useBoolean/useBoolean.ts
```diff
@@ -20,15 +20,19 @@ type UseBooleanReturn = {
...
+ * @throws Will throw an error if `defaultValue` is an invalid boolean value.
  * @public
  * @see [Documentation](https://usehooks-ts.com/react-hook/use-boolean)
  * @example
  * ```tsx
  * const { value, setTrue, setFalse, toggle } = useBoolean(true);
  * ```
  */
-export function useBoolean(defaultValue?: boolean): UseBooleanReturn {
-  const [value, setValue] = useState(!!defaultValue)
```
luckrnx09, line 31: Invoid to use `!!`, It reduces readability.

juliencrn (approved):
Good 👍

PR #584, opened 2024-04-07
Allow useLocalStorage mutation callbacks to be called in render phase

jraoult:
The callbacks returned by `useLocalStorage` are currently "event callbacks" and, therefore, can't be called in the render phase. It is in contradiction with the recommended approach to adjust state (cf. https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) where it should be possible to do:

```tsx
function App() {
  const [lsCount, setLsCount] = useLocalStorage('lsCount', 0);

  if (lsCount === 0) {
    // this should work but fails right now
    setLsCount(1);
  }
[...]
```

This PR removes this limitation.

### DIFF
--- a/packages/usehooks-ts/src/useLocalStorage/useLocalStorage.test.ts
+++ b/packages/usehooks-ts/src/useLocalStorage/useLocalStorage.test.ts
@@ -208,6 +208,22 @@ describe('useLocalStorage()', () => {
     expect(result.current[1] === originalCallback).toBe(true)
   })
 
+  it('setValue can be called in render phase', () => {
+    const { result } = renderHook(() => {
+      const [value, setValue] = useLocalStorage('count', 1)
+
+      // Adjust state with the recommended approach
+      // cf. https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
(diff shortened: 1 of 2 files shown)


jraoult:
@juliencrn as promised here's the PR for https://github.com/juliencrn/usehooks-ts/issues/549

jraoult:
@juliencrn ping just in case you missed it 😉

jraoult:
@juliencrn have you had the time to take a look at tis PR?

ddanielcruzz:
First of all thanks for this awesome project @juliencrn

I was wondering if there any updates regarding merging this?

kelkes:
I would love to have this merge... :)

jraoult:
@kelkes I might be wrong, but I think the commit history of this project doesn't paint a bright future for this library: https://github.com/juliencrn/usehooks-ts/commits/master/, so I would not hold my breath. Avoid it if you have just started using it, or migrate to another solution when possible. If you are only interested in this particular hook, https://usehooks.com/uselocalstorage might be better.

kelkes:
@jraoult i guess you are right... thanks for the alternative.

PR #610, opened 2024-06-27
fix: fix useDebounceCallback isPending logic

PeterChen1997:
I found a problem about the isPending response. It will return true after the first call.

I think it should return false after there is no pending call👀

Here is the demo: use the fixed hook, and the response may be correct

https://stackblitz.com/edit/vitejs-vite-kbwkhh?file=src%2FApp.tsx

### DIFF
--- a/packages/usehooks-ts/src/useDebounceCallback/useDebounceCallback.ts
+++ b/packages/usehooks-ts/src/useDebounceCallback/useDebounceCallback.ts
@@ -77,26 +77,39 @@ export function useDebounceCallback<T extends (...args: any) => ReturnType<T>>(
   options?: DebounceOptions,
 ): DebouncedState<T> {
   const debouncedFunc = useRef<ReturnType<typeof debounce>>()
+  const isPendingRef = useRef(false)
 
+  const parsedFunc = useCallback(
+    (...args: Parameters<T>) => {
+      const result = func(...args)
+      isPendingRef.current = false
+      return result
+    },
+    [func]
+  )
+  
   useUnmount(() => {
     if (debouncedFunc.current) {
       debouncedFunc.current.cancel()
+      isPendingRef.current = false
     }
   })
 
   const debounced = useMemo(() => {
-    const debouncedFuncInstance = debounce(func, delay, options)
+    const debouncedFuncInstance = debounce(parsedFunc, delay, options)
 
     const wrappedFunc: DebouncedState<T> = (...args: Parameters<T>) => {
+      isPendingRef.current = true
       return debouncedFuncInstance(...args)
     }
 
     wrappedFunc.cancel = () => {
       debouncedFuncInstance.cancel()
+      isPendingRef.current = false
     }
 
     wrappedFunc.isPending = () => {
-      return !!debouncedFunc.current
+      return isPendingRef.current
     }
 
     wrappedFunc.flush = () => {
@@ -104,12 +117,12 @@ export function useDebounceCallback<T extends (...args: any) => ReturnType<T>>(
     }
 
     return wrappedFunc
-  }, [func, delay, options])
+  }, [parsedFunc, delay, options])
 
   // Update the debounced function ref whenever func, wait, or options change
   useEffect(() => {
-    debouncedFunc.current = debounce(func, delay, options)
-  }, [func, delay, options])
+    debouncedFunc.current = debounce(parsedFunc, delay, options)
+  }, [parsedFunc, delay, options])
 
   return debounced
 }


packages/usehooks-ts/src/useDebounceCallback/useDebounceCallback.ts
```diff
@@ -77,39 +77,52 @@ export function useDebounceCallback<T extends (...args: any) => ReturnType<T>>(
...
     return wrappedFunc
-  }, [func, delay, options])
+  }, [parsedFunc, delay, options])
 
   // Update the debounced function ref whenever func, wait, or options change
   useEffect(() => {
-    debouncedFunc.current = debounce(func, delay, options)
-  }, [func, delay, options])
+    debouncedFunc.current = debounce(parsedFunc, delay, options)
+  }, [parsedFunc, delay, options])
```
alexeychikk, line 125: I don't seem to understand how this hook's cancellation works.
We are debouncing `debouncedFunc` and `debounced` independently from each other, and return `debounced`. 
So when component unmounts, `debouncedFunc` is cancelled and not `debounced`.

PeterChen1997: Oh I got it, fixed it with the true ref 🤣

packages/usehooks-ts/src/useDebounceCallback/useDebounceCallback.ts
```diff
@@ -77,39 +77,52 @@ export function useDebounceCallback<T extends (...args: any) => ReturnType<T>>(
   options?: DebounceOptions,
 ): DebouncedState<T> {
   const debouncedFunc = useRef<ReturnType<typeof debounce>>()
+  const isPendingRef = useRef(false)
 
+  const parsedFunc = useCallback(
+    (...args: Parameters<T>) => {
+      const result = func(...args)
+      isPendingRef.current = false
```
simon-lammes, line 85: Would it make sense to put this line into a `finally` block, so that it even works as expected if `func` throws an exception?

PeterChen1997: wow, that's coolll, fixed!

packages/usehooks-ts/src/useDebounceCallback/useDebounceCallback.ts
simon-lammes: I cannot see the purpose of `debouncedFunc` at all. It seems redundant since all operations could just be called on the debounced function created with `useMemo`.

The call to `debouncedFunc.current.cancel()` doesn't have the expected effect. When testing this hook, you can see that pending invocations will eventually execute even if the component has been unmounted. While this isn't the behavior I anticipated, it is reasonable. As a consumer, I would expect the debounced function to be cancelled (or flushed) only if I explicitly opt-in via a public, documented API.

Therefore, I believe this call (along all other usages of `debouncedFunc`) can be safely removed, as it has no effect and its removal simplifies the code without losing any necessary functionality.

[Here](https://github.com/simon-lammes/usehooks-ts/blob/ad64f88bddaad0dd4ea656d2daed34885dbab3f6/packages/usehooks-ts/src/useDebounceCallback/useDebounceCallback.ts#L80) is an example where I removed `debouncedFunc` altogether. However, I cancelled working on my PR because I figured that my change of cancelling or flushing on unmount would be an undesirable, breaking change. So please ignore the use of `useUnmount` and the added `flushOnUnmount` option.

PeterChen1997: Thanks for your detailed review! That's a honor for me lol.

But I checked the cancel behavior is correct in this demo. https://stackblitz.com/edit/vitejs-vite-nwfbcy?file=src%2FTest.tsx

While you unmount the component, the debounced function just canceled. You can have a look in the demo console.

I think that's a reasonable behavior for the react users. Otherwise it maybe cause some memory leak or setState warning 👀

PeterChen1997:
@simon-lammes is this ok for merge 👀

simon-lammes (approved):

simon-lammes (approved):

simon-lammes:
@PeterChen1997 I don't see any issues with the propsed changes. However, I don't have the permission to merge.

PeterChen1997:
> @PeterChen1997 I don't see any issues with the propsed changes. However, I don't have the permission to merge.

Pretty thx!

@alexeychikk please take a look if you have time ~

alexeychikk:
> > @PeterChen1997 I don't see any issues with the propsed changes. However, I don't have the permission to merge.
> 
> Pretty thx!
> 
> @alexeychikk please take a look if you have time ~

I'm not a contributor/maintainer. I just wanted to copy implementation in my project but found out it's poopy :)  
Anyway, it looks okay now :)

alexeychikk (approved):

PeterChen1997:
> > > @PeterChen1997 I don't see any issues with the propsed changes. However, I don't have the permission to merge.
> > 
> > 
> > Pretty thx!
> > @alexeychikk please take a look if you have time ~
> 
> I'm not a contributor/maintainer. I just wanted to copy implementation in my project but found out it's poopy :) Anyway, it looks okay now :)

wow thx too~

@juliencrn juliencrn please have a look ~

PeterChen1997:
ping for notice ~

PR #620, opened 2024-07-24
feat: custom initial step

axelrindle:
Adds a second parameter `initialStep: number = 1` to the `useStep` hook which allows the user to define a custom initial step between `1` and `maxStep`.


Darijusch:
i would like to see this in usehooks as well, could you cover your new initialStep with unit tests?

PR #628, opened 2024-08-06
Add option 'flushOnUnmount' to useDebounceCallback

simon-lammes:
# Motivation/Purpose

The `useDebounceCallback` hook is suitable for auto-save functionality. However, the current behavior is that pending invocations will be canceled when the component unmounts. Consequently, if the user leaves a page while there is a pending save invocation, those changes will be lost.

By setting 'flushOnUnmount' to true, pending changes will be saved as soon as the user leaves the page.

# API changes

The new option has the following _non-breaking_ API definition:

If set to true, a pending invocation of the debounced function will be immediately invoked when the component unmounts. If not set or set to false, a pending invocation will be canceled and therefore not invoked when the component unmounts.

# Additional changes

- Fixed `isPending` not returning the correct value as filed in #610
- I removed the variable `debouncedFunc` because of the following assumptions (I am not 100 percent certain and I might have overlooked something):
  - We called lodash's debounce in two places, thereby getting two debounced functions (`debounced` and `debouncedFunc`) which can be confusing and lead to errors.
  - `debounced` seems to be the one "actually being used" while `debouncedFunc` was only used for the "erroneous" `isPending` behavior
  - Therefore, the call `debouncedFunc.current.cancel()` in `useUnmount` did not cancel the "right" debounced function. Therefore, it does not take any effect. I tested this in a "real" application: When you console.log in your callback and leave the page, you will see that the console.log will eventually be called even though your component has already unmounted. Although replicating this behavior is straightforward, it is outside the primary focus of this PR. However, I can create a demo if you think it would be beneficial.

# Checklist

- [x] The new API has been tested with 2 test cases.
- [x] The fixed isPending behavior has been tested with 2 test cases.
- [x] The new API option is documented.

### DIFF
--- a/packages/usehooks-ts/src/useDebounceCallback/useDebounceCallback.test.ts
+++ b/packages/usehooks-ts/src/useDebounceCallback/useDebounceCallback.test.ts
@@ -107,4 +107,109 @@ describe('useDebounceCallback()', () => {
     // The callback should be invoked immediately after flushing
     expect(debouncedCallback).toHaveBeenCalled()
   })
+
+  it('should flush pending invocations on unmount if configured to do so', () => {
+    const delay = 500
+    const debouncedCallback = vitest.fn()
+    const { result, unmount } = renderHook(() =>
+      useDebounceCallback(debouncedCallback, delay, {
+        // This option is being tested.
+        flushOnUnmount: true,
+      }),
+    )
+
+    act(() => {
+      result.current('argument')
+    })
+
+    // The callback should not be invoked immediately
+    expect(debouncedCallback).not.toHaveBeenCalled()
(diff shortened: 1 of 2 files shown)


simon-lammes:
I just realized that this pull request is _breaking._

The existing implementation made me think that the current behavior is to "cancel on unmount". However, as noted above, the behavior is to neither cancel nor flush on unmount. I don't know whether that is by design or mistake.

Either way, I will have to close this PR as it is breaking.

Maybe, one can come up with a nice, non-breaking solution for all 3 cases:
- neither flush nor cancel on unmount (seemingly current behavior)
- cancel on unmount
- flush on unmount

PR #629, opened 2024-08-13
Feature: add useThrottle hook

jiwooproity:
Adds a new hook called useThrottle that executed after a certain period of time based on the last function among the functions that are executed quickly.

## Todo
- [x] useThrottle hook
- [x]  Docs
- [x]  Example
- [x]  Tests


packages/usehooks-ts/src/useThrottle/useThrottle.ts
```diff
@@ -0,0 +1,33 @@
...
+ * // OR
+ * const throttle = useThrottle(); // Initial value is default value (1s)
+ * // Use throttle function in your component, Only one of the events that are processed redundantly can be processed.
+ * ```
+ */
+
+export function useThrottle(
+  defaultInterval = 1000,
+): (callback: () => void, interval?: number) => void {
+  const ref = useRef<NodeJS.Timeout>()
```
minecrawler, line 22: This should use `@types/web` instead of node types. Even the demo runs in the browser, so no NodeJS in sight ;)

JuditKaramazov (approved):

jiwooproity:
@juliencrn Hello.
I implemented throttle hook.
If you think you need to add this hook, Can you review this hook?

Sicria:
This would be incredibly helpful, along with a `useThrottleCallback` to match the `useDebounceCallback` function.

bstaruk:
+1 on the previous comments  -- this hook would be a great addition.

jiwooproity:
@Sicria @bstaruk 
Thank you, I will be happy if this hook helps someone.
I hope this feature gets approved!

pedrokohler:
@juliencrn are we merging this any time soon?

PR #636, opened 2024-09-17
fix(dark-mode): avoid localStorage overwrite by system dark mode status

qwqcode:
## Problem Description

While using `useDarkMode`, I noticed that after refreshing the page, the value saved in `localStorage` does not take effect (observing that `localStorage` is always overwritten by the current system's dark mode status), even though I set the `initializeWithValue` option to true (the documentation states, "If true (default), the hook will initialize by reading `localStorage`").

## Cause of the Issue

The value saved in `localStorage` does not take effect on the next page load because the value retrieved by `const isDarkOS = useMediaQuery()` immediately **overwrites** the result read from `useLocalStorage()` when the `useDarkMode` hook initializes (since `useIsomorphicLayoutEffect` executes immediately). This behavior is unexpected, causing the functionality of saving the dark mode state to `localStorage` to fail.

## Fix

Add `const allowDarkOSChange = useIsMounted()`. Before the page is mounted, prevent applying the dark mode result from `isDarkOS`, rather than immediately executing and overwriting the value from `localStorage`.

Test cases have been added to verify the existence of this issue, and the problem was confirmed. The code modification has passed the tests successfully.

### DIFF
--- a/packages/usehooks-ts/src/useDarkMode/useDarkMode.test.ts
+++ b/packages/usehooks-ts/src/useDarkMode/useDarkMode.test.ts
@@ -124,4 +124,29 @@ describe('useDarkMode()', () => {
 
     expect(result.current.isDarkMode).toBe(false)
   })
+
+  it('should update dark mode when OS preference changes', () => {
+    const { updateMatches } = mockMatchMedia(false)
+    const { result, rerender } = renderHook(() => useDarkMode())
+    expect(result.current.isDarkMode).toBe(false)
+    updateMatches(true)
+    rerender() // trigger `useIsomorphicLayoutEffect`
+    expect(result.current.isDarkMode).toBe(true)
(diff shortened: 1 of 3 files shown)


adamwdennis (approved):
LGTM

adamwdennis:
@qwqcode this needs a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)

JamesWatling (approved):
Yes this helps so much!

qwqcode:
> @qwqcode this needs a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)

Got it, I added it.

nikischin:
This works for me. Can we get this merged @juliencrn ?

Thank you so much!

PR #666, opened 2025-01-01
feature: add new hook useescClose for closing element in keypress

jihostudy:
This PR introduces a new custom hook, `useEscClose`.

`useEscClose` is designed to close open pages or components, such as modals, by pressing the ESC key. Users often expect to be able to close open elements like modals or toasts using the ESC key. This hook provides a solution for such scenarios. By passing a boolean value (indicating whether the target is open) and a callback function (to be executed after closing) as arguments, this hook attaches an ESC event listener to the `window` object and handles the ESC key action accordingly.

### DIFF
--- a/packages/usehooks-ts/src/index.ts
+++ b/packages/usehooks-ts/src/index.ts
@@ -31,3 +31,5 @@ export * from './useTimeout'
 export * from './useToggle'
 export * from './useUnmount'
 export * from './useWindowSize'
+
+export * from './useEscclose'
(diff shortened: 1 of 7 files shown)


A1X5H04:
## Suggestion

This would be much better if the hook supports **hotkeys** instead of specific key, in this case an 'ESC' key.

For eg: 
If a user wants to do something when a key is pressed, you can expose a callback function that will execute when a key (or certain key combination) is pressed, you can use any helper library or create one from scratch, you may also include a threshold when to press key combination etc

jihostudy:
> , you can use any helper library or create one from scratch, you may also include a threshold when to press key combination

Thank you for your comment. As you mentioned, it seems a good idea to address various key scenarios in addition to the ESC key.

PR #665, opened 2024-12-22
Bugfix/#602 - Add support for React 19

iwan-uschka:
This PR adds support for React 19 by fixing TS issues regarding `useRef` (see https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument).

To be able to use React 19 throughout the project i had to update some other dependencies as well.

To be able to use the TypeScript version of this project in VS Code i removed all `typescript` dependencies from all `package.json` and added one single `typescript` dependency to the root `package.json`. I hope this is ok.

Before this change, they were two versions of `typescript`: `5.4.3` and `5.3.3`. Now we have only one version of `typescript`: `5.4.3`. To avoid running into warnings like

```bash
usehooks-ts:lint: =============
usehooks-ts:lint: 
usehooks-ts:lint: WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
usehooks-ts:lint: 
usehooks-ts:lint: You may find that it works just fine, or you may not.
usehooks-ts:lint: 
usehooks-ts:lint: SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.4.0
usehooks-ts:lint: 
usehooks-ts:lint: YOUR TYPESCRIPT VERSION: 5.4.3
usehooks-ts:lint: 
usehooks-ts:lint: Please only submit bug reports when using the officially supported version.
usehooks-ts:lint: 
usehooks-ts:lint: =============
```

when calling `pnpm lint` i had to update `@typescript-eslint` to `7.18.0`. This required some fixes to add to the codebase.

PR #670, opened 2025-01-21
fix: documentation sidebar component to show on mobile

### LINKED ISSUE #664: Documentation site sidebar does not show on mobile
### Describe the bug

When visiting the documentation site, the sidebar is not visible, meaning documentation is not viewable without entering landscape or desktop mode.

### To Reproduce

View any documentation page on mobile device. Such as: https://usehooks-ts.com/introduction

### Expected behavior

Sidebar is either already showing or can be opened with a button.

### Additional context

_No response_

junnjiee:
Fixes #664 (There was no way to access the `LeftSidebar` component in mobile view, which contained links to each hook documentation)

`LeftSidebar` wasn't able to show on mobile as `hidden md:block` was applied on the component itself

I changed the component to take in an extra `className` prop, so that only selected `LeftSidebar` components will be hidden.

I also removed the Features and Documentation nav links in the `MobileNav` component - it was weird having the Documentation and Introduction link together, since it led to the same page.

Edited the marketing layout so that `LeftSidebar` is able to be shown on the home page too

Before:
image

After:
image

### DIFF
--- a/.changeset/cyan-plums-cheat.md
+++ b/.changeset/cyan-plums-cheat.md
@@ -1,0 +1,5 @@
+---
+"www": patch
+---
+
+Fix LeftSidebar component not showing on mobile
--- a/apps/www/src/app/(docs)/layout.tsx
+++ b/apps/www/src/app/(docs)/layout.tsx
@@ -19,7 +19,7 @@ export default async function DocsLayout({ children }: DocsLayoutProps) {
     <>
       <header className="sticky top-0 z-40 w-full border-b bg-background">
         <div className="container flex h-16 items-center space-x-4 sm:justify-between sm:space-x-0">
-          <MainNav items={docsConfig.mainNav}>
+          <MainNav>
             <LeftSidebar items={docsConfig.sidebarNav} hooks={hooks} />
           </MainNav>
           <div className="flex flex-1 items-center space-x-4 sm:justify-end">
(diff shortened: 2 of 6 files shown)


iwan-uschka:
Thanks @junnjiee16 for creating that PR. I am new to this project in terms of contributing to it but i would like to challenge the changes you've made because the main navigation on the homepage now differs between mobile and tablet/desktop view.

I would suggest to
- remove the main navigation on the homepage completely because it doesn't add real value
  - `Documentation` and `Explore the docs` link to the same URL and are initially visible even on smaller screens
  - `Features` contains an anchor link to a block that is initially visible even on smaller screens
- apply your changes to all pages but the root/home page

iwan-uschka:
In https://github.com/juliencrn/usehooks-ts/pull/665 we try to figure out if it makes sense to work on PRs in this repository at all because it seems to be dead. But who knows. Anyway: if you want to make this PR work you have to add a changeset. I went through the same process couple of weeks ago :) Feel free to scroll through https://github.com/juliencrn/usehooks-ts/pull/665 to get some information about it.

junnjiee16:
Hi @iwan-uschka, thank you for the suggestions. I committed changes to remove the main navigation links from the page as well as include a patch changeset

>  apply your changes to all pages but the root/home page

For this suggestion, I couldn't find a way to do it. Since `LeftSidebar` is used in `layout.tsx` itself, I have to edit this page to add the `hidden` class so that the `LeftSidebar` component in desktop view will be hidden.

PR #676, opened 2025-02-07
Update dev dependencies

iwan-uschka:
Consolidate `typescript` dependency versions to a single one referenced in the root `package.json`.  Before this change, they were two versions of TypeScript: 5.4.3 and 5.3.3. Now we have only one version of TypeScript: 5.4.3. To avoid running into warnings like

```
usehooks-ts:lint: =============
usehooks-ts:lint: 
usehooks-ts:lint: WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
usehooks-ts:lint: 
usehooks-ts:lint: You may find that it works just fine, or you may not.
usehooks-ts:lint: 
usehooks-ts:lint: SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.4.0
usehooks-ts:lint: 
usehooks-ts:lint: YOUR TYPESCRIPT VERSION: 5.4.3
usehooks-ts:lint: 
usehooks-ts:lint: Please only submit bug reports when using the officially supported version.
usehooks-ts:lint: 
usehooks-ts:lint: =============
```

when calling pnpm lint i had to update @typescript-eslint to 7.18.0. This required some fixes to add to the codebase.

I also added VS Code setting to prefer TypeScript version of workspace in IDE instead of the default one of VS Code itself.

### DIFF
--- a/packages/usehooks-ts/src/useIntersectionObserver/useIntersectionObserver.demo.tsx
+++ b/packages/usehooks-ts/src/useIntersectionObserver/useIntersectionObserver.demo.tsx
@@ -28,7 +28,7 @@ export default function Component() {
   return (
     <>
       {Array.from({ length: 5 }).map((_, index) => (
-        <Section key={index + 1} title={(index + 1).toFixed()} />
+        <Section key={index + 1} title={(index + 1).toString()} />
       ))}
     </>
   )
--- a/packages/usehooks-ts/src/useIntersectionObserver/useIntersectionObserver.demo.tsx
+++ b/packages/usehooks-ts/src/useIntersectionObserver/useIntersectionObserver.demo.tsx
@@ -28,7 +28,7 @@ export default function Component() {
   return (
     <>
       {Array.from({ length: 5 }).map((_, index) => (
-        <Section key={index + 1} title={`${index + 1}`} />
+        <Section key={index + 1} title={(index + 1).toFixed()} />
       ))}
     </>
   )
--- a/packages/usehooks-ts/src/useScrollLock/useScrollLock.test.ts
+++ b/packages/usehooks-ts/src/useScrollLock/useScrollLock.test.ts
@@ -108,7 +108,9 @@ describe('useScrollLock()', () => {
 
     const scrollbarWidth = window.innerWidth - document.body.scrollWidth
 
-    expect(document.body.style.paddingRight).toBe(`${scrollbarWidth}px`)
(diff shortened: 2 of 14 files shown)


packages/usehooks-ts/src/useIntersectionObserver/useIntersectionObserver.demo.tsx
```diff
@@ -28,7 +28,7 @@ export default function Component() {
   return (
     <>
       {Array.from({ length: 5 }).map((_, index) => (
-        <Section key={index + 1} title={`${index + 1}`} />
+        <Section key={index + 1} title={(index + 1).toFixed()} />
```
lachieh: Best to use `.toString()` here. This would cause a difference in output. Template literals coerce their values to strings, which calls the `.toString()` method instead of `.toFixed()`

```js
let x = 123.456
x.toString() // "123.456"
x.toFixed() // "123"
```

```suggestion
        <Section key={index + 1} title={(index + 1).toString()} />
```

packages/usehooks-ts/src/useScrollLock/useScrollLock.test.ts
```diff
@@ -108,7 +108,9 @@ describe('useScrollLock()', () => {
 
     const scrollbarWidth = window.innerWidth - document.body.scrollWidth
 
-    expect(document.body.style.paddingRight).toBe(`${scrollbarWidth}px`)
+    expect(document.body.style.paddingRight).toBe(
+      `${scrollbarWidth.toFixed()}px`,
```
lachieh: Same as previous comment

packages/usehooks-ts/src/useScrollLock/useScrollLock.ts
```diff
@@ -90,7 +90,7 @@ export function useScrollLock(
           0
 
         const scrollbarWidth = offsetWidth - target.current.scrollWidth
-        target.current.style.paddingRight = `${scrollbarWidth + currentPaddingRight}px`
+        target.current.style.paddingRight = `${(scrollbarWidth + currentPaddingRight).toFixed()}px`
```
lachieh: And here

lachieh (approved):
Small change regarding string coercion but looks good to me. @juliencrn will have to give final approval since they own it but I figure it doesn't hurt to have extra eyes on.

lachieh (changes requested):
Should probably have marked for changes

PR #679, opened 2025-02-14
fix: incorrect persistence useDarkMode

### LINKED ISSUE #512: useDarkMode resets

tihonove:
## Fix useDarkMode persistence

This PR fixes #512. This pull request addresses the issue with the `useDarkMode` hook in what I believe is the best way: without modifying other hooks or adding overhead, unlike other PRs that attempt to fix this bug.

### Changes:

1. If there is no value in `localStorage`, it will not appear if the OS switches to dark mode.
2. The absence of a value in `localStorage` means it will correspond to the OS setting or the `defaultValue`.
3. Other hooks are used naturally without any modifications.

This fix ensures that the `useDarkMode` hook persistence correctly and efficiently.

### DIFF
--- a/.changeset/five-papayas-pull.md
+++ b/.changeset/five-papayas-pull.md
@@ -1,0 +1,5 @@
+---
+"usehooks-ts": patch
+---
+
+fix useDarkMode persistence
--- a/packages/usehooks-ts/src/useDarkMode/useDarkMode.test.ts
+++ b/packages/usehooks-ts/src/useDarkMode/useDarkMode.test.ts
@@ -124,4 +124,20 @@ describe('useDarkMode()', () => {
 
     expect(result.current.isDarkMode).toBe(false)
   })
+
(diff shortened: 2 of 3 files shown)


TimESQuit:
When will `usehooks-ts` make a new release that includes this change? It has been more than 8 months so far.

tihonove:
I've gived up and remove my fokr. I think that `useDarkMode` is buggy by desing, so the right way to get manage with dark mode is using [useTernaryDarkMode](https://usehooks-ts.com/react-hook/use-ternary-dark-mode) function of this lib.

PR #680, opened 2025-02-26
Change the type of refs for react 19

### LINKED ISSUE #663: [BUG] useOnClickOutside doesn't allow null ref
### Describe the bug

The sample code in https://usehooks-ts.com/react-hook/use-on-click-outside#usage throws a TypeError:
```
Argument of type 'RefObject<HTMLDivElement | null>' is not assignable to parameter of type 'RefObject<HTMLElement> | RefObject<HTMLElement>[]'.
  Type 'RefObject<HTMLDivElement | null>' is not assignable to type 'RefObject<HTMLElement>'.
    Type 'HTMLDivElement | null' is not assignable to type 'HTMLElement'.
      Type 'null' is not assignable to type 'HTMLElement'.ts(2345)
```

### To Reproduce

```
  const ref = useRef(null)

  const handleClickOutside = () => {
    // Your custom logic here
    console.log('clicked outside')
  }

  const handleClickInside = () => {
    // Your custom logic here
    console.log('clicked inside')
  }

  useOnClickOutside(ref, handleClickOutside)
```

### Expected behavior

Should accept null

### Additional context

_No response_

elee1766:
see https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument

so now 
`MutableRef is now deprecated in favor of a single RefObject type which useRef will always return:`
```
interface RefObject<T> {
  current: T
}

declare function useRef<T>: RefObject<T>

```
and
`useRef still has a convenience overload for useRef<T>(null) that automatically returns RefObject<T | null>. To ease migration due to the required argument for useRef, a convenience overload for useRef(undefined) was added that automatically returns RefObject<T | undefined>.
`

so i think a change like this is correct? i think it fixes #663 as well

### DIFF
--- a/packages/usehooks-ts/src/useEventListener/useEventListener.ts
+++ b/packages/usehooks-ts/src/useEventListener/useEventListener.ts
@@ -31,7 +31,7 @@ function useEventListener<
   handler:
     | ((event: HTMLElementEventMap[K]) => void)
     | ((event: SVGElementEventMap[K]) => void),
-  element: RefObject<T>,
+  element: RefObject<T | null | undefined>,
   options?: boolean | AddEventListenerOptions,
 ): void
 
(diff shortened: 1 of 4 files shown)


grundmanise:
@juliencrn - Sorry to ping, but I wanted to check if this `usehooks-ts` package is still being maintained, considering how many PRs remain unmerged.

elee1766:
> @juliencrn - Sorry to ping, but I wanted to check if this `usehooks-ts` package is still being maintained, considering how many PRs remain unmerged.

fwiw I gave up and switched to mantine hooks

PR #708, opened 2025-11-12
feat: add a faster mode to useLocalStorage via `priority` flag

quantizor:
I have a use case where a local storage setting directly affects page layout and I'd like to try and reduce the page jumpiness as much as possible during initial load.

### DIFF
--- a/.changeset/hot-bats-report.md
+++ b/.changeset/hot-bats-report.md
@@ -1,0 +1,7 @@
+---
+'usehooks-ts': major
+---
+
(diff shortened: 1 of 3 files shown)


quantizor:
Hmm doesn't seem to make much of a difference from experimentation -- up to you if you want to accept or close this.

emmsdan:
> Hmm doesn't seem to make much of a difference from experimentation -- up to you if you want to accept or close this.

I agree not much difference, I think the little difference still matters

PR #720, opened 2026-04-10
fix: cancel correct debounced instance on unmount in useDebounceCallback

haunc08:
## Summary

The `useUnmount` cleanup in `useDebounceCallback` was cancelling a separate debounced instance (created in `useEffect`) instead of the actual instance used by callers (created in `useMemo`). This meant pending debounced calls were never cancelled on unmount.

## Changes

- Store the `debouncedFuncInstance` from `useMemo` in the ref directly, so `useUnmount` cancels the correct instance
- Remove the redundant `useEffect` that created a separate unused debounced instance
- Add unit test verifying pending callbacks are cancelled on unmount


vascoteixeira-aurum (approved):

umanghome:
Nice. The root cause is causing an issue for us too when running vitest. Looking forward to this being merged.

